PR #64, opened 2021-12-07
[QU-390] Sequel patch for message threading

HoonBaek:
## For Internal Contributors

[QU-390](https://sendbird.atlassian.net/browse/QU-390)

## Description Of Changes

- Move the location of message status and timestamp into the message content
- Use paragraph tag in the Text message and OG message contents and set their word-break property to break-word
- Add type definition with `replyType` to the App and Channel components props

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -387,6 +387,7 @@ interface ChannelProps {
   renderChatHeader?: (props: RenderChatHeaderProps) => React.ReactNode;
   renderUserProfile?: (props: RenderUserProfileProps) => React.ReactNode;
   queries?: ChannelQueries;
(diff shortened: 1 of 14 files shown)


src/index.d.ts
```diff
@@ -448,6 +449,7 @@ interface AppProps {
     resizingWidth?: number | string,
     resizingHeight?: number | string,
   };
+  replyType?: ReplyType;
```
sravan-s, line 452: 👍

sravan-s (approved):

PR #94, opened 2022-02-17
fix: localization parameter for some format calls

sravan-s:
localization parameter was not passed correctly to some
date-fn calls. Which caused some strings to be still in English

Refs:
* https://sendbird.atlassian.net/browse/UIKIT-1536
* https://sendbird.atlassian.net/browse/SBISSUE-7849


HoonBaek (approved):
LGTM

PR #98, opened 2022-02-19
Fix: Jest expects commonjs module by default

aburgel:
I'm running into two issues with a create react app v5 app and sendbird-uikit. I was able to fix them and clean up a few other parts of the build process.

The first issue occurs when running jest tests. Jest seems to expect `node_modules` to already be transpiled (or whatever it's called), but your package.json points it's main entrypoint at the esm variant, even though there is a commonjs variant produced.

This is what the error looks like. The fix is to update `main` in package.json to point to the commonjs index.js.
```
    /Users/aburgel/src/hh-frontend/node_modules/sendbird-uikit/release/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export { default as SendBirdProvider } from './SendbirdProvider.js';
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

      1 | import React, { FC } from "react";
    > 2 | import { SendBirdProvider } from "sendbird-uikit";
```

The second issue is it that the index.css sourcemap produces an error. TBH, I'm not sure why this is happening, but I was able to fix it.
```
WARNING in ./node_modules/sendbird-uikit/dist/index.css (./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[1].oneOf[5].use[1]!./node_modules/postcss-loader/dist/cjs.js??ruleSet[1].rules[1].oneOf[5].use[2]!./node_modules/source-map-loader/dist/cjs.js!./node_modules/sendbird-uikit/dist/index.css)
Module Warning (from ./node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/Users/aburgel/src/hh-frontend/node_modules/sendbird-uikit/dist/index.css.map' file: Error: ENOENT: no such file or directory, open '/Users/aburgel/src/hh-frontend/node_modules/sendbird-uikit/dist/index.css.map'
 @ ./node_modules/sendbird-uikit/dist/index.css 8:6-232 22:17-24 26:7-21 58:25-39 59:36-47 59:50-64 63:6-73:7 64:54-65 64:68-82 70:42-53 70:56-70 72:21-28 83:0-202 83:0-202 84:22-29 84:33-47 84:50-64 61:4-74:5
 @ ./src/components/SendBirdWrapper.tsx 10:0-39
 @ ./src/App.tsx 8:0-57 28:45-60
 @ ./src/index.tsx 8:0-24 23:35-38
```


package.json
```diff
@@ -2,12 +2,11 @@
   "name": "sendbird-uikit",
   "version": "2.6.0",
   "description": "React based UI kit for sendbird",
-  "main": "release/index.js",
-  "style": "release/dist/index.css",
-  "typings": "release/index.d.ts",
-  "module": "release/index.js",
+  "main": "dist/cjs/index.js",
```
aburgel, line 5: This fixes the jest issue.

I also moved the files into dist. I'm not sure why there was `release` and `dist` but it didn't seem necessary and was easy to clean up.

This should be backwards compatible since users shouldn't be referencing the `release` paths directly.

package.json
```diff
@@ -19,9 +18,8 @@
   "scripts": {
     "test": "jest",
     "test-update-snapshot": "jest -u",
-    "cp-css": "./scripts/copy-css.sh",
```
aburgel, line 22: I cleaned up the rollup config so that this is no longer necessary. Rollup now outputs everything into `dist`.

rollup.config.js
```diff
@@ -97,6 +99,7 @@ module.exports = ({
           },
         ],
       ],
+      babelHelpers: 'bundled',
```
aburgel, line 102: This fixes a warning:

```
babelHelpers: 'bundled' option was used by default. It is recommended to configure this option explicitly, read more here: https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers
```

rollup.config.js
```diff
@@ -35,14 +35,16 @@ module.exports = ({
   },
   output: [
     {
-      dir: 'release/dist/cjs',
+      dir: 'dist/cjs',
       format: 'cjs',
       sourcemap: true,
+      exports: 'named'
```
aburgel, line 41: This fixes a warning:

```
(!) Entry module "src/lib/Sendbird.jsx" is implicitly using "default" export mode, which means for CommonJS output that its default export is assigned to "module.exports". For many tools, such CommonJS output will not be interchangeable with the original ES module. If this is intended, explicitly set "output.exports" to either "auto" or "default", otherwise you might want to consider changing the signature of "src/lib/Sendbird.jsx" to use named exports only.
```

sravan-s:
Hey, thanks for the contribution, we need some time to a look at this~~ Probably take care of it end of this week/early next week

package.json
sravan-s: That was due to some legacy customer request, but, can remove now :)

aburgel:
@sravan-s any chance of landing this soon? it's blocking us from an upgrade since without it our jest-based test suite has trouble compiling.

sravan-s:
@aburgel We had soem holidays and bunch of other releases here. But we included it in  2.7.0 beta. We will test for a while and see if there are anyone with issues and release 2.7.0 formally

https://github.com/sendbird/sendbird-uikit-react/releases/tag/v2.7.0-0

PR #100, opened 2022-02-21
Feature: [v3] Provide auto select flag in channel list

HoonBaek:
## For Internal Contributors

[UIKit-1503](https://sendbird.atlassian.net/browse/UIKit-1503)

## Description Of Changes

* Add an autoSelectChannelItem props to the App and ChannelList component
* Call the first onChannelSelect event handler only when autoSelectChannelListItem props is true
* Add a "checkout" command to reset node_modules easily

## Types Of Changes

- [ ] Bugfix
- [x] New feature
- [ ] Build configuration


HoonBaek:
This change causes conflict with QA fixes, so I will apply this change in the QA PR instead here.

HoonBaek:
https://github.com/sendbird/sendbird-uikit-react/pull/101

PR #102, opened 2022-02-23
feat: List exports into a file

sravan-s:
Done:
* List exports into a file
* chore: Do not commit size-snapshot.json

Todo:
* Add types to all exported files

Ticket: https://sendbird.atlassian.net/browse/UIKIT-1510


HoonBaek (approved):
I agree with changing to OpenChannel instead of Openchannel

PR #108, opened 2022-03-02
Fix: [v2] Appear MessageStatus component in Safari

HoonBaek:
## For Internal Contributors

[UIKIT-1577](https://sendbird.atlassian.net/browse/UIKIT-1577)

## Description Of Changes

* Fix the issue that inline-flex causes disappearing component in Safari

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/ui/MessageStatus/index.scss
+++ b/src/ui/MessageStatus/index.scss
@@ -3,19 +3,18 @@
 .sendbird-message-status {
   position: relative;
   display: inline-block;
-  flex-direction: row;
   width: 100%;
   height: 100%;
 
   .sendbird-message-status__icon {
     position: relative;
     display: inline-block;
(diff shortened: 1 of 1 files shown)


src/ui/MessageStatus/index.scss
```diff
@@ -2,19 +2,19 @@
 
 .sendbird-message-status {
   position: relative;
-  display: inline-flex;
+  display: inline-block;
   flex-direction: row;
```
sravan-s, line 6: flex-direction: row; doesnt work with display: inline-block; no?

HoonBaek: Right, let me fix

sravan-s (approved):

PR #124, opened 2022-03-29
Fix: [v3] Activate scroll event on the message search list

HoonBaek:
## For Internal Contributors

[UIKIT-1626](https://sendbird.atlassian.net/browse/UIKIT-1626)

## Description Of Changes

* Activate scroll event on the message search list component

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/smart-components/MessageSearch/index.scss
+++ b/src/smart-components/MessageSearch/index.scss
@@ -95,6 +95,6 @@
     box-sizing: border-box;
     width: 100%;
     height: 100%;
-    overflow-y: scroll;
+    overflow-y: hidden;
   }
 }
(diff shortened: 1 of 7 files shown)


src/smart-components/MessageSearch/index.scss
```diff
@@ -95,6 +95,6 @@
     box-sizing: border-box;
     width: 100%;
     height: 100%;
-    overflow-y: scroll;
+    overflow-y: hidden;
```
sravan-s, line 98: What happened here?

sravan-s (approved):

src/smart-components/MessageSearch/index.scss
HoonBaek: The outer component is set by scroll, so the internal component wasn't applied scroll

PR #128, opened 2022-04-12
Fix lint warnings and add types

sravan-s:
* Add types to index.d.ts
* Fix warnings when bundled library was used with external projects


src/smart-components/ChannelList/context/ChannelListProvider.tsx
```diff
@@ -279,6 +279,6 @@ function useChannelListContext (): ChannelListProviderInterface {
 
 export {
   ChannelListProvider,
-  ChannelListContext,
+  // ChannelListContext,
```
HoonBaek, line 282: Why don't we remove this line?

HoonBaek (approved):

PR #143, opened 2022-05-03
fix: Add rate limit to markAsDelivered call

sravan-s:
fixes: https://sendbird.atlassian.net/browse/UIKIT-1718


HoonBaek (approved):
Oh this is the same fix with https://github.com/sendbird/sendbird-uikit-react/pull/144/files 
but for v2. thanks

PR #148, opened 2022-05-18
use disableAutoSelect from ChannelListProvider props in initial state

neilfulwiler:
## External Contributions

and we'll evaluate if we can setup a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) to allow for the contribution.

## For Internal Contributors

[<TICKET_ID>](https://sendbird.atlassian.net/browse/<TICKET_ID>)

## Description Of Changes

Add a brief description of the changes that you have involved in this PR

## Types Of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/smart-components/ChannelList/context/ChannelListProvider.tsx
+++ b/src/smart-components/ChannelList/context/ChannelListProvider.tsx
@@ -160,7 +160,7 @@ const ChannelListProvider: React.FC<ChannelListProviderProps> = (props: ChannelL
 
   const [channelListStore, channelListDispatcher] = useReducer(
     channelListReducers,
-    channelListInitialState,
+    {...channelListInitialState, disableAutoSelect: disableAutoSelect !== undefined ? disableAutoSelect : channelListInitialState.disableAutoSelect}
   ) as [ChannelListStoreInterface, CustomUseReducerDispatcher];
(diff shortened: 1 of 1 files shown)


sravan-s:
Hi @neilfulwiler can you share the context of the fix?

neilfulwiler:
Hey @sravan-s! Yes of course, after submitting I saw the big block about external contributions, but realize it would help to provide context anyway.

Basically if you specify disableAutoSelect in the ChannelListProvider that doesn't actually propagate down into the channelListReducer (which seems to have its own disableAutoSelect state although this state is never mutated), which means that disableAutoSelect effectively doesn't work right now for the ChannelListProvider. This fix just pushes through that flag into the initialState used with the reducer so it should fix what we're seeing

sravan-s:
@neilfulwiler Thanks for the contribution..

About the open source contribution guidelines, bug fixes like these are fine. Its usually the larger features contributions that we might not accept

HoonBaek:
Thank you @neifulwiler I modified this issue and it's a hundred percent thanks to you 🙏
(I fixed it to fit our convention) https://github.com/sendbird/sendbird-uikit-react/pull/152

PR #153, opened 2022-05-24
fix: add type defn for useChannelList

sravan-s:


src/index.d.ts
```diff
@@ -497,6 +497,7 @@ declare module '@sendbird/uikit-react/ChannelList' {
 declare module '@sendbird/uikit-react/ChannelList/context' {
   export type ChannelListProvider = React.FunctionComponent<ChannelListProviderProps>;
   export function useChannelListContext (): ChannelListProviderInterface;
```
HoonBaek, line 499: Don't need to remove the previous hook? or are we going to do it all in one time

HoonBaek (approved):

PR #156, opened 2022-06-03
feat: Mention improvements

sravan-s:
* Show profile on clicking a mention
* Visual highlight when user is mention

Fixes:
https://sendbird.atlassian.net/browse/UIKIT-1786
https://sendbird.atlassian.net/browse/UIKIT-1787

### DIFF
--- a/src/ui/MentionLabel/index.tsx
+++ b/src/ui/MentionLabel/index.tsx
@@ -10,7 +10,6 @@ import type SendBird from 'sendbird';
 
 import ContextMenu, { MenuItems } from '../ContextMenu';
 import Label, { LabelTypography, LabelColors } from '../Label';
-import { UserProfileContext } from '../../lib/UserProfileContext';
 import UserProfile from '../UserProfile';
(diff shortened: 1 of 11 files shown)


src/ui/MentionLabel/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    [sdk, mentionedUserId],
+  );
+  return (
+    <ContextMenu
+      menuTrigger={(toggleDropdown: () => void): ReactElement => (
+        <a
+          className={`
+            sendbird-word__mention
+            ${iAmBeingMentioned && 'sendbird-word__mention--me'}
```
HoonBaek, line 59: How about this
```
${iAmBeingMentioned ? 'sendbird-word__mention--me' : ''}
```
because above returns 'false' as string type

HoonBaek (approved):
LGTM, just one suggestion here

src/ui/MentionLabel/index.tsx
sravan-s: yes will do 👍

PR #164, opened 2022-06-30
Bugfix [v3] QM-1547_Keep suggested mention max count

HoonBaek:
## For Internal Contributors

[QM-1547](https://sendbird.atlassian.net/browse/QM-1547)
[QM-1552](https://sendbird.atlassian.net/browse/QM-1552)
[QM-1553](https://sendbird.atlassian.net/browse/QM-1553)

## Description Of Changes

* QM-1547
  * Keep the maximum suggestion count even current user is included in the fetch
* QM-1552
  * Differentiate getting nickname and userId from stringObj
  * Fetch the `User` using userId instead of the nickname
  * Display mention text using the nickname
* QM-1553
  * Enable sending 1:1 message on the user profile of the user mentioned text

So the nickname could be changed, but userId is unchangeable

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/ui/UnknownMessageItemBody/index.tsx
+++ b/src/ui/UnknownMessageItemBody/index.tsx
@@ -2,13 +2,14 @@ import React, { ReactElement, useContext } from 'react';
 import './index.scss';
 
 import Label, { LabelTypography, LabelColors } from '../Label';
-import { CoreMessageType, getClassName } from '../../utils';
+import { getClassName } from '../../utils';
 import { LocalizationContext } from '../../lib/LocalizationContext';
+import { FileMessage, UserMessage } from '@sendbird/chat/message';
 
 interface Props {
   className?: string | Array<string>;
   isByMe?: boolean;
-  message: CoreMessageType;
+  message: UserMessage | FileMessage | any;
   mouseHover?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


src/ui/UnknownMessageItemBody/index.tsx
```diff
@@ -2,13 +2,14 @@ import React, { ReactElement, useContext } from 'react';
...
-import { CoreMessageType, getClassName } from '../../utils';
+import { getClassName } from '../../utils';
 import { LocalizationContext } from '../../lib/LocalizationContext';
+import { FileMessage, UserMessage } from '@sendbird/chat/message';
 
 interface Props {
   className?: string | Array<string>;
   isByMe?: boolean;
-  message: CoreMessageType;
+  message: UserMessage | FileMessage | any;
```
sravan-s, line 12: can we exclude `any;`

sravan-s (approved):

src/ui/UnknownMessageItemBody/index.tsx
HoonBaek: Then how about using `unknown`?

HoonBaek: Let me merge this PR first, we can change in the next PR

PR #168, opened 2022-06-30
fix: add parent class to openChannelSettings

sravan-s:
fixes: https://sendbird.atlassian.net/browse/QM-1567


HoonBaek (approved):
Thanks a lot

PR #171, opened 2022-07-01
fix: operator ui

sravan-s:
Operator UI was also broken, figured out after participant fix
fixes: https://sendbird.atlassian.net/browse/QM-1567


HoonBaek (approved):
Really appreciate!!

PR #212, opened 2022-07-12
fix: add node_module polyfills by defualt

sravan-s:
Before: need to add stream and buffer libraries manually to webapck5 projects


HoonBaek (approved):
LGTM

PR #214, opened 2022-07-18
fix: Type definition for channellist and setting

sravan-s:


HoonBaek (commented):
It could be there are more places that the previous name is being used

sravan-s:
Checked and there werent any I could find

HoonBaek (approved):
Just check the three files that I said

PR #215, opened 2022-07-19
feat: accept customApiHost & customWebSocketHost

sravan-s:
Accept customApiHost & customWebSocketHost on SendbirdProvider and App
These urls let the SDK to connect to custom domains

fixes: https://sendbird.atlassian.net/browse/UIKIT-1853

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -17,6 +17,8 @@
     appId: string;
     userId: string;
     accessToken?: string;
+    customApiHost?: string,
+    customWebSocketHost?: string,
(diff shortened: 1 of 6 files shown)


HoonBaek (approved):
LGTM, however, I have two ideas
1. I think it's good to talk with the other platforms and try to keep in sync with them because it's our interface
2. I think `apiHost` and `webSocketHost` is enough for names because we have used `accessTocken`, so isn't it fine to not use prefix `custom`

sravan-s:
1. Told other teams, they will add it too
2. just using same name from SDK

PR #250, opened 2022-08-10
Fix: Update SendableMessage to UserMessage and FileMessage

HoonBaek:
## For Internal Contributors

[UIKIT-2055](https://sendbird.atlassian.net/browse/UIKIT-2055)

## Description Of Changes

Update SendableMessage to UserMessage and FileMessage

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/src/lib/selectors.ts
+++ b/src/lib/selectors.ts
@@ -572,11 +572,11 @@ export const getUpdateUserMessage = (state: SendBirdState) => (
  *  channel: GroupChannel | OpenChannel,
  *  message: SendableMessage,
  * )
- *  .then(() => {})
+ *  .then((deletedMessage) => {})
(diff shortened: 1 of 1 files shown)


src/lib/selectors.ts
```diff
@@ -586,7 +586,7 @@ export const getDeleteMessage = (state: SendBirdState) => (
             topics.DELETE_MESSAGE,
             { messageId, channel },
           );
-          resolve(message);
+          resolve();
```
sravan-s, line 589: We dont need to pass down the deleted message?

HoonBaek: we don't need to, because there's no message of response

sravan-s (approved):

PR #253, opened 2022-08-11
fix: add missing type defns from production file

sravan-s:
fixes: 
https://sendbird.atlassian.net/browse/UIKIT-2061


HoonBaek (approved):
Appriciate for these

PR #270, opened 2022-08-16
fix: move @sendbird/uikit-react as build dependency

sravan-s:
@sendbird/chat got bundled separately by rollup after v3
This causes inheritance trees to break

fixes: https://sendbird.atlassian.net/browse/UIKIT-1988


HoonBaek (approved):

sravan-s:
We remove this commit from main ~~

PR #275, opened 2022-08-26
fix: Change type of react elements

HoonBaek:
## For Internal Contributors

[UIKIT-2053](https://sendbird.atlassian.net/browse/UIKIT-2053)

## Description Of Changes

* Change every `React.ReactNode` and `React.Component` to `React.ReactElement`
* Use the type of SendbirdError
* Use the type MessageSearchQueryParams
* Use enum MessageSearchOrder.TIMESTAMP in the message search query params instead of `'ts' as const`

**ReactNode** could be `string | number | null | undefined | ReactElement | portal` and this(expecting string or number) causes **warning** when we use it like `<CustomComp />`
```typescript
// in the component
{ renderMessage } = props
const CustomMessage = useMemo(() => {
  return renderMessage({ ... });
}, []);
return (
  <div>
    <CustomMessage />
  </div>
);
```
so expecting **ReactElement** is better for our case

## Types Of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Build configuration

### DIFF
--- a/CHANGELOG_v2.md
+++ b/CHANGELOG_v2.md
@@ -294,7 +294,7 @@ channel OpenChannelSettings
 - Features/Bugs:
   * ChannelList
     * disableUserProfile: boolean
-    * renderUserProfile: React.Component
+    * renderUserProfile: React.ReactElement
 
 ## 1.2.4(Sept 17, 2020) Deprecated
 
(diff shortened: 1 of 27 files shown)


CHANGELOG_v2.md
```diff
@@ -294,7 +294,7 @@ channel OpenChannelSettings
 - Features/Bugs:
   * ChannelList
     * disableUserProfile: boolean
-    * renderUserProfile: React.Component
+    * renderUserProfile: React.ReactElement
 
 ## 1.2.4(Sept 17, 2020) Deprecated
```
sravan-s, line 300: Is better not to do this, because in v2, its still React.Component
And we make this change in v3.1.2/3

sravan-s (approved):

HoonBaek:
let's leave this pending until versioning policy is set up

HoonBaek:
https://github.com/sendbird/sendbird-uikit-react/pull/277
we will use ReactNode and ReactElement for custom render props

PR #282, opened 2022-09-06
feat: allow customers to use rem units for typography

sravan-s:
Rem is better unit for accessibility support/mobile
So, we are adding rem as unit for typography/font size
We expect font-size 16px as base font-size, but to avoid breaking change,
customers should pass `config.isREMUnitEnabled` prop to `SendbirdProvider` to use "rem" units

ticket: https://sendbird.atlassian.net/browse/UIKIT-2176

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -77,7 +77,7 @@ declare module "SendbirdUIKitGlobal" {
       maxMentionCount?: number,
       maxSuggestionCount?: number,
     };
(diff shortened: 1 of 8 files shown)


HoonBaek (commented):
Why don't we choose some name starting with `is~~` instead of `use~~`?
I remember we decided to follow that way

sravan-s:
sure can do that

HoonBaek (approved):

PR #162, opened 2022-06-30
Bugfix [v3] QA for Chat v4 migration

HoonBaek:
## For Internal Contributors

[QM-1510](https://sendbird.atlassian.net/browse/QM-1510)
[QM-1516](https://sendbird.atlassian.net/browse/QM-1516)
[QM-1542](https://sendbird.atlassian.net/browse/QM-1542)
[QM-1518](https://sendbird.atlassian.net/browse/QM-1518)

## Description Of Changes

QM-1510
* Use the params `coverImage` instead of `coverUrl` when updating channel cover image

QM-1516
QM-1542
* Add a handler for onUserBanned

QM-1518
* Set up the storybook mention preference to the default

The other stuff
* Translate useHandleChannelEvents to typescript
* Apply optional chaining for many channel instances
* Update many types with new Sendbird chat types

## Types Of Changes

What types of changes does your code introduce to this project?
Put an `x` in the boxes that apply_

- [x] Bugfix
- [ ] New feature
- [ ] Documentation (correction or otherwise)
- [ ] Cosmetics (whitespace, appearance (ex) Prettier)
- [ ] Build configuration
- [x] Improvement (refactor code)

PR #301, opened 2022-09-21
Fix: allMessages data not correct when specific condition

mango906:
History of this PR: https://github.com/sendbird/sendbird-uikit-react/pull/162#discussion_r953531091

I checked that external PR is not prepared, but please understand that I am requesting PR because you asked me to sent it if PR exists. 🙏 

## Description Of Changes

Hi! I am facing one issue while using your product. For the data of messages in the channel, i am using the messages array, which is the return value of the `useChannelContext` hook.

```typescript
const { messages } = useChannelContext;
```

In normal case, the `messages` array data is valid, but in special case, the `messages` array data is not valid. The special case I'm talking about is If it is not the first channel in the channel list, the messages received in real time are not reflected in the messages array. 
(It is reproduced equally when moving another channel and returning to the first channel. I think "channel shift" is the key word on this issue.)

```javascript
/* Normal case */

messages // ['Hi One', 'Hi Two'] (example)

// Recieve `Hi Three` message

// expected
messages // ['Hi One', 'Hi Two', 'Hi Three'] 
// result
messages // ['Hi One', 'Hi Two', 'Hi Three'] 

/* Special cases mentioned above */

messages // ['Hi One', 'Hi Two'] (example)

// Recieve `Hi Three` message

// expected
messages // ['Hi One', 'Hi Two', 'Hi Three'] 
// result
messages // ['Hi One', 'Hi Two'] 

```

So, when I looked at the code, it was confirmed that the `hasMoreNext` value was only true when special conditions I said.
Therefore, the value of the messages array was not valid because the logic below was not being executed because it was being early-return. When I removed check hasMoreNext condition logic, I was able to confirm that it worked as I intended.

But I don't understand your code perfectly, so I'm not sure what side effects it will have. If there is a side effect caused by the change, it will need to be corrected and If there is a better solution, it would be better to follow it. I just want the problem I'm facing to be solved. 

I would appreciate it if you could consider the problem 🙏  Thanks!

## Types Of Changes

- [x] Bugfix

### DIFF
--- a/src/smart-components/Channel/context/ChannelProvider.tsx
+++ b/src/smart-components/Channel/context/ChannelProvider.tsx
@@ -282,7 +282,7 @@ const ChannelProvider: React.FC<ChannelContextProps> = (props: ChannelContextPro
 
   // Hook to handle ChannelEvents and send values to useReducer using messagesDispatcher
   useHandleChannelEvents(
-    { currentGroupChannel, sdkInit, hasMoreNext },
+    { currentGroupChannel, sdkInit },
     {
       messagesDispatcher,
       sdk,
--- a/src/smart-components/Channel/context/hooks/useHandleChannelEvents.ts
+++ b/src/smart-components/Channel/context/hooks/useHandleChannelEvents.ts
@@ -18,8 +18,7 @@ import * as messageActions from '../dux/actionTypes';
  * sdkInit: bool
  */
 interface DynamicParams {
-  sdkInit: boolean;
-  hasMoreNext: boolean;
+  sdkInit: boolean;  
   currentGroupChannel: GroupChannel;
 }
 interface StaticParams {
@@ -32,8 +31,7 @@ interface StaticParams {
 
 function useHandleChannelEvents(
   {
-    sdkInit,
-    hasMoreNext,
+    sdkInit,    
     currentGroupChannel,
   }: DynamicParams,
   {
@@ -49,9 +47,8 @@ function useHandleChannelEvents(
     const channelHandlerId = uuidv4();
     if (channelUrl && sdkInit) {
       const channelHandler: GroupChannelHandler = {
-        onMessageReceived: (channel, message) => {
-          // Do not update when hasMoreNext
-          if (compareIds(channel?.url, channelUrl) && !hasMoreNext) {
+        onMessageReceived: (channel, message) => {          
+          if (compareIds(channel?.url, channelUrl)) {
             let scrollToEnd = false;
             try {
               const { current } = scrollRef;


sravan-s:
@mango906 Thanks for the PR, let us check this, will merge in 2-3 days
Honestly, we dont mind bug fixes, its feature requests that's what we usually wont take
Mainly because, we have to have parity with design team, RN, ios and android

mango906:
> @mango906 Thanks for the PR, let us check this, will merge in 2-3 days Honestly, we dont mind bug fixes, its feature requests that's what we usually wont take Mainly because, we have to have parity with design team, RN, ios and android

Thanks :)

hellfreya:
@mango906  감사합니다!

mango906:
> @mango906 감사합니다!

요 이슈는 빠르게 고쳐주시면 감사하겠습니다 🙇

hellfreya:
추가로, 해당 이슈는 재현이 되지 않아서 검증이 되지 않는데요. 추가적인 재현 환경에 대한 정보를 주실 수 있으실까요 ?

mango906:
> 추가로, 해당 이슈는 재현이 되지 않아서 검증이 되지 않는데요. 추가적인 재현 환경에 대한 정보를 주실 수 있으실까요 ?

@hellfreya 

채팅 어플리케이션을 만들고 있는데, `useChannelContext()` 을 사용해 `allMessages` 데이터를 받아오고, 해당 데이터를 렌더링하면서 채팅 화면을 그리고 있습니다. 만약 사용자가 첫번쨰로 입장한 Channel이라면, `allMessages` 데이터가 정상적으로 잘 내려옵니다.

하지만 사용자가 다른 Channel로 이동하거나, 다른 Channel로 이동했다가 처음 Channel로 다시 돌아오게 되면, allMessages의 데이터가 정확하지 않습니다. 웹소켓 통신을 통해 데이터를 주고 받는거처럼 보이는데, 앞의 재현상황에서 allMessages에 새로 받게 된 메시지가 들어가지 않습니다. (새로고침하면 데이터가 정상적으로 노출됩니다.) 

혹시 제가 어렵게 설명한 부분이 있다면 지적해주시면 감사하겠습니다 🙇

hellfreya:
네 전달드린뒤 추가 피드백이 필요하면 알려드리겠습니다

HoonBaek:
Hi @mango906 

image
`hasMoreNext` should be included in the condition, because we block adding incoming messages to the real UI list while we fetch the message list from the middle of the entire message list.
It might be fetched from the middle when you use the message search feature and the scroll moved

And I tried to reproduce this issue following the above steps, but I couldn't reproduce this bug on my side, if you leave the specific environment (version, browser) and attach a video, it will help me to reproduce it.

mango906:
Hi @HoonBaek I will attach the video together to help you understand.
I simply created a cra project and recreated the problem situation. 
And I sent a message to other connected applications, called `new message` and `new message2`

environment
* @sendbird/uikit-react: "^3.2.1",
* browser: Chrome v105


In the first video, you can see that the message `new message` comes well.
But In the second video, however, `new message2` message is reflected in the Channel List, but not in the Channel.
As mentioned earlier, in the second video, the channel was moved.
I think the `allMessages` from `useChannelContext` hooks is not valid. 


https://user-images.githubusercontent.com/38802280/194227322-e1838209-c7c7-4c30-bad5-80c39cdef4ef.mov

https://user-images.githubusercontent.com/38802280/194227328-c1758670-1806-4ba7-91a4-6f8b86735b56.mov

mango906:
Would it be easier to reproduce if I attached the code for the demo project?

PR #304, opened 2022-09-22
fix: use queries from @sendbird/chat

sravan-s:
Use imported versions of GroupChannelListQueryParams and ApplicationUserListQueryParams


HoonBaek (approved):
LGTM

PR #307, opened 2022-09-22
fix: do not use connectionHandler when there is sessionHandler

sravan-s:
ConnectionHandler seems to invalidate sessionHandler


HoonBaek (approved):
This prop `configureSession` is not on our docs,
we need to update the docs with the other missed info too, asap

sravan-s:
Is there in the handler PR

PR #320, opened 2022-09-27
revert: bundle @sendbird/chat as peer dependency

sravan-s:
Even though it is more desirable,
Adding `@sendbird/chat` as peer dependency fails the bundled library 
Need to investigate more


HoonBaek (approved):

bang9:
image
Can you try this? @sravan-s @HoonBaek

PR #340, opened 2022-10-20
feat: Responsive App component

sravan-s:
* App component is now responsive
* MediaQuery accepts boolean value
  Use case: Customer can detect the devices of their preference and setup mobile view
  instead of only relying on width


src/smart-components/App/AppLayout.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react';
```
sravan-s, line 1: Not too enthusiastic about exporting layouts

PR #342, opened 2022-10-23
feat: BottomSheet style menu for Messages

sravan-s:
interface BottomSheetProps {
  className?: string;
  children: React.ReactElement;
  onBackdropClick?: () => void;
}


src/ui/MobileMenu/MobileContextMenu.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import type { MobileContextMenuProps } from './types';
+
+// import {
```
sravan-s, line 4: Will implement later in the PR

PR #350, opened 2022-11-01
feat: add new props isLoading to ChannelUI

sravan-s:
It is better to have seperate prop to show loader in channel,
becuase, other customers might have used empty/null to show invalid channel
and it would be a breaking change for them

Usage:
```
<Channel channelUrl={currentChannelUrl} isLoading={!currentChannelUrl} />
```## External Contributions

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -473,6 +473,7 @@ declare module "SendbirdUIKitGlobal" {
   };
 
   export interface ChannelUIProps {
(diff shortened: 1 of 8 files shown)


sravan-s:
After: Kapture 2022-11-02 at 06 48 04

sravan-s:
Before:
Kapture 2022-11-02 at 06 50 36

HoonBaek (approved):

PR #353, opened 2022-11-02
fix: add a channel when user is invited

sravan-s:
fixes: https://sendbird.atlassian.net/browse/UIKIT-1744


src/smart-components/ChannelList/utils.js
```diff
@@ -34,6 +34,17 @@ const createEventHandler = ({
...
+      if (isInvited) {
+        channelListDispatcher({
+          type: channelActions.USER_INVITED,
+          payload: channel,
+        });
+      }
+    },
     onUserBanned: (channel, user) => {
       const { currentUser } = sdk;
       logger.info('Channel | useHandleChannelEvents: onUserBanned', channel);
```
HoonBaek, line 50: Can you change this log to this?
```suggestion
      logger.info('Channel: onUserBanned', channel);
```

Only this logger has different format here

HoonBaek (approved):
LGTM

PR #361, opened 2022-11-09
fix: use ref instead of querySelector for DOM manipulation

sravan-s:
QuerySelector is inaccurate when multiple channels are rendered parallel

This doesnt include root for dropdowns/fileviewer etc

### DIFF
--- a/src/smart-components/Channel/context/utils.js
+++ b/src/smart-components/Channel/context/utils.js
@@ -8,19 +8,19 @@ import { OutgoingMessageStates } from '../../../utils/exports/getOutgoingMessage
 const UNDEFINED = 'undefined';
 const { SUCCEEDED, FAILED, PENDING } = getSendingMessageStatus();
 
-export const scrollIntoLast = (intialTry = 0) => {
+export const scrollIntoLast = (intialTry = 0, scrollRef) => {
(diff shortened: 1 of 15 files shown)


src/smart-components/Channel/context/utils.js
```diff
@@ -8,19 +8,19 @@ import { OutgoingMessageStates } from '../../../utils/exports/getOutgoingMessage
 const UNDEFINED = 'undefined';
 const { SUCCEEDED, FAILED, PENDING } = getSendingMessageStatus();
 
-export const scrollIntoLast = (intialTry = 0) => {
+export const scrollIntoLast = (intialTry = 0, scrollRef) => {
```
sravan-s, line 11: @HoonBaek  use this pattern in threading

HoonBaek: Looks good, thank you 🙏

HoonBaek: variable name should be `initialTry`?

HoonBaek (approved):
Looks good

src/smart-components/Channel/context/utils.js
sravan-s: si... Will apply before merging

PR #372, opened 2022-12-01
fix: create channel sometimes had empty operatorID

sravan-s:
use sendbird state to access currentUserID and use it incase prop value is empty
Also, remove legacy HOC pattern


src/smart-components/App/index.jsx
```diff
@@ -67,7 +67,7 @@ export default function App(props) {
       isReactionEnabled={isReactionEnabled}
       isMentionEnabled={isMentionEnabled}
       onUserProfileMessage={(channel) => {
-        currentChannel(channel);
+        setCurrentChannel(channel);
```
HoonBaek, line 70: 🙏

PR #406, opened 2023-02-20
feat: options to render icons on MessageInput

sravan-s:
Add props:
```
  renderFileUploadIcon?: () =>  React.ReactElement;
  renderVoiceMessageIcon?: () =>  React.ReactElement;
  renderSendMessageIcon?: () =>  React.ReactElement;
```

To Channel & Thread
> Important: renderVoiceMessageIcon and renderSendMessageIcon only customizes IconUI
But, renderFileUploadIcon helps you customize/extend functionality
For example: it can be extended to open camera, open a new modal to drag and drop files, convert files before upload ~

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
   "files": [
     "release/**/*",
     "dist/**/*",
-    "package-lock.json",
     "LICENSE",
     "CHANGELOG.md"
   ],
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -9,7 +9,6 @@
   "files": [
     "release/**/*",
     "dist/**/*",
-    "package-lock.json",
```
sravan-s, line 12: it is ignored anyways, also this file goes nowhere ~
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

```
Conversely, some files are always ignored:
...
package-lock.json (use [npm-shrinkwrap.json](https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json) if you wish it to be published)
```

HoonBaek (approved):
👍

PR #407, opened 2023-02-21
feat: add prop `activeChannelUrl` to ChannelList

sravan-s:
Add prop: activeChannelUrl 

activeChannelUrl gives customer an option to pragmatically set a channel from a parent component/ router

If channel is present in client side, it is used
If not present in client side, channel is fetched from backend and added to end of the list and set as selected
If not present, no errors will be thrown

If user clicks other channel, the item set from outside is removed and clicked channel will be selected 

If activeChannelUrl is present, ChannelList behaves similar to "disableAutoSelect" is true

example
```
const MyChat = () => {
  const {my_active_channel} = useRouter();
  return (
    <ChannelList activeChannelUrl={my_active_channel} />
  );
```
fixes: https://sendbird.atlassian.net/browse/UIKIT-3180

### DIFF
--- a/src/smart-components/ChannelList/context/ChannelListProvider.tsx
+++ b/src/smart-components/ChannelList/context/ChannelListProvider.tsx
@@ -33,6 +33,7 @@ import { CustomUseReducerDispatcher } from '../../../lib/SendbirdState';
 import channelListReducers from '../dux/reducers';
 import channelListInitialState from '../dux/initialState';
 import { CHANNEL_TYPE } from '../../CreateChannel/types';
+import useActiveChannelUrl from './hooks/useActiveChannelUrl';
 
 interface ApplicationUserListQuery {
   limit?: number;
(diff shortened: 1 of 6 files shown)


src/smart-components/ChannelList/context/ChannelListProvider.tsx
```diff
@@ -150,10 +152,13 @@ const ChannelListProvider: React.FC<ChannelListProviderProps> = (props: ChannelL
     onBeforeCreateChannel,
     sortChannelList,
     overrideInviteUser,
-    disableAutoSelect,
+    activeChannelUrl,
```
HoonBaek, line 155: Isn't it breaking change or our public interface?
We should take one of the ways
1. Apply this change in the next minor version up & check Docs & notice to SE team & remove disableAutoSelect props from type definition
2. Keep both props

sravan-s: no, its not... I import it two lines below.. is an augmentation ~

sravan-s: `const disableAutoSelect = props?.disableAutoSelect || !!activeChannelUrl;`
Is a trick to use have same behavior if customer sets  an activeChannelUrl ~

HoonBaek (approved):

PR #438, opened 2023-03-16
feat: mention should work on paste

sravan-s:
Mentions should be preserved from copy paste from sendbird-messages
and message input

* Make sure you are posting mentions of users from same channel
* We dont support pasting of rich text from other applications
* For copying simple text, we recommend using paste option in message context-menu

Conditions tested:
1. paste simple text
2. paste text with mention
3. paste text with mention and text
4. paste text with mention and text and paste again before and after
5. copy message with mention(only one mention, no other text) and paste
6. copy message with mention from input and paste(before and after)

fixes: https://sendbird.atlassian.net/browse/UIKIT-2980


HoonBaek:
LGTM, thanks @sravan-s

PR #449, opened 2023-03-23
feat: replace renderToString(react-dom) with custom fn

sravan-s:
Replace renderToString from react-dom/server with custom function
This function was creating issue in customers with cra@4 & react@17
fixes: https://sendbird.atlassian.net/browse/UIKIT-3403

### DIFF
--- a/src/ui/MentionUserLabel/renderToString.ts
+++ b/src/ui/MentionUserLabel/renderToString.ts
@@ -7,17 +7,13 @@ type renderToStringParams = {
 };
 
 export default function renderToString({ userId, nickname }: renderToStringParams): string {
-  const el = `
-    <span
-      contenteditable='false'
-      data-userid="${userId}"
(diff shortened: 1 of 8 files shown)


src/ui/MessageInput/hooks/usePaste/insertTemplate.tsx
```diff
@@ -1,23 +1,14 @@
 import React from 'react';
```
HoonBaek, line 1: Need to remove this line

src/ui/MentionUserLabel/renderToString.ts
```diff
@@ -0,0 +1,23 @@
+// cretes a sanitized string from a mention user label
```
HoonBaek, line 1: I guess `MentionUserLabel` would be better to search later

HoonBaek (approved):

PR #450, opened 2023-03-24
fix: menu position in tight screens

sravan-s:
* condition where some menus get clipped in left side:
  * usually user profile in channel moderation
* context menu of last item in channel gets clipped under

fixes 1. & 3. from: https://sendbird.atlassian.net/browse/SBISSUE-10883

### DIFF
--- a/src/ui/ContextMenu/MenuItems.tsx
+++ b/src/ui/ContextMenu/MenuItems.tsx
@@ -17,6 +17,9 @@ interface MenuItemsState {
   handleClickOutside: (e: MouseEvent) => void;
 }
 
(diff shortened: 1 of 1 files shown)


sravan-s:
before:
Screen Shot 2023-03-24 at 3 04 01 PM
Screen Shot 2023-03-24 at 3 04 13 PM

sravan-s:
after:
Screen Shot 2023-03-24 at 2 50 31 PM
Screen Shot 2023-03-24 at 2 50 43 PM

HoonBaek (approved):

HoonBaek:
> Especially for supporting mobile size screen

PR #451, opened 2023-03-24
[v3.4.3] (Mar 24 2023)

sravan-s:
Features:
* Add rollup-plugin-size-snapshot for bundle-size Run rollup-plugin-size-snapshot on build, we will check bundle size before every release
* Move old samples to use vite React team these days are using vite for their samples, CRA is discourged
* Run code coverage on commenting `./coverage` Check code coverage on PR comment
* Add prop to disable Channel & Thread inputs Add prop: `disabled?: false` for Channel & Thread MessageInputWrapper
* Replace renderToString(react-dom) with custom fn Replace renderToString from react-dom/server with custom function This function was creating issue in customers with cra@4 & react@17

Fixes:
* Replace outdated CSS rules `justify-content: start;` and  `height: fill-available;`
* Menu position in tight screens
  * Condition where some menus get clipped in left side: * Usually user profile in channel moderation
  * Context menu of last item in channel gets clipped in the bottom

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sendbird/uikit-react",
-  "version": "3.4.2",
+  "version": "3.4.3",
   "description": "React based UI kit for sendbird",
   "main": "dist/index.js",
   "style": "dist/index.css",
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,36 +1,36 @@
 {
   "index.js": {
-    "bundled": 7133,
-    "minified": 6731,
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sendbird/uikit-react",
-  "version": "3.4.2",
+  "version": "3.4.3-rc.1",
```
HoonBaek, line 3: need to remove `rc` before release

sravan-s: will do ~

PR #459, opened 2023-03-28
fix: clean up the interface of thread module

HoonBaek:
### Description Of Changes

Issue
* Had to cover parent message with brackets before putting it to the Thread. So it caused confusing work around

Fix
* Check if the message is parent message or thread message inside of the Thread
  * Use case (Before)
  ```javascript
  <Thread
    message={message}
    // or
    message={{ parentMessage }}
  />
  ```
  * Use case (After)
  ```javascript
  <Thread
    message={message}
    // or
    message={parentMessage}
  />
  ```
* Remove props on the interface: `includePollDetails`
* Add util functions `isParentMessage` and `isThreadMessage`

[UIKIT-3325](https://sendbird.atlassian.net/browse/UIKIT-3325)

[UIKIT-3325]: https://sendbird.atlassian.net/browse/UIKIT-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/smart-components/Thread/context/ThreadProvider.tsx
+++ b/src/smart-components/Thread/context/ThreadProvider.tsx
@@ -3,7 +3,7 @@ import { User } from '@sendbird/chat';
 import { GroupChannel } from '@sendbird/chat/groupChannel';
 import { BaseMessage, FileMessage, FileMessageCreateParams, UserMessage } from '@sendbird/chat/message';
 
-import { getNicknamesMapFromMembers, isThreadMessage } from './utils';
+import { getNicknamesMapFromMembers, getParentMessageFrom } from './utils';
 import { UserProfileProvider } from '../../../lib/UserProfileContext';
 import { CustomUseReducerDispatcher } from '../../../lib/SendbirdState';
 import useSendbirdStateContext from '../../../hooks/useSendbirdStateContext';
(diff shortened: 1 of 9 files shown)


src/smart-components/Thread/context/__test__/utils.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  parentMessage: mockParentMessage,
+  parentMessageId: 1,
+  threadInfo: null,
+};
+
+describe('Thread/utils - isParentMessage', () => {
+  it('should comfirm if messages are parent message', () => {
+    expect(isParentMessage(mockParentMessage as UserMessage)).toBe(true);
+    expect(isParentMessage(mockThreadMessage as UserMessage)).toBe(false);
+  });
```
sravan-s, line 26: true and false should be seperate assertions, same for `Thread/utils - isThreadMessage `

HoonBaek: That means,, it needs to be divided into two test?

src/smart-components/Thread/context/ThreadProvider.tsx
```diff
@@ -73,6 +72,8 @@ export const ThreadProvider: React.FC<ThreadProviderProps> = (props: ThreadProvi
     renderUserProfile,
     onUserProfileMessage,
   } = props;
+  const propsMessage = props?.message;
+  const propsParentMessage = isThreadMessage(propsMessage) ? propsMessage?.parentMessage : propsMessage;
```
sravan-s, line 76: is good to make a small function(not hook) and add a unit test,
or.. comment why we do this :)

sravan-s (approved):

PR #466, opened 2023-03-31
fix: Modify condition of checking parent and thread messages

HoonBaek:
### Description Of Changes

* Do not check threadInfo to improve the parent message, because every normal messages potentially can be a parent message
* The type of messageId should be the number

### DIFF
--- a/src/smart-components/Thread/context/utils.ts
+++ b/src/smart-components/Thread/context/utils.ts
@@ -25,15 +25,14 @@ export const getParentMessageFrom = (message: UserMessage | FileMessage): UserMe
 export const isParentMessage = (message: UserMessage | FileMessage): boolean => {
   return (
     message?.parentMessage === null
-    && message?.parentMessageId === null
-    && message?.threadInfo !== null
+    && !message?.parentMessageId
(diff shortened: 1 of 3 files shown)


src/smart-components/Thread/context/utils.ts
```diff
@@ -25,15 +25,14 @@ export const getParentMessageFrom = (message: UserMessage | FileMessage): UserMe
...
-    && message?.threadInfo !== null
+    && !message?.parentMessageId
   );
 };
 
 export const isThreadMessage = (message: UserMessage | FileMessage): boolean => {
   return (
     message?.parentMessage !== null
-    && message?.parentMessageId !== null
+    && message?.parentMessageId
```
sravan-s, line 35: add unit test before merge pls

sravan-s (approved):

src/smart-components/Thread/context/utils.ts
HoonBaek: It's sure that the type of parentMessaeId is number, so I wanted to check if it's 0 or more

PR #470, opened 2023-04-03
fix: Rename useMessageScrollRef to messageScrollRef

angelhcp:
## External Contributions

I'm proposing a variable renaming.

The current variable name is: `useMessageScrollRef`. This is not ideal, because React treats it as a custom hook. Since it is not a custom hook, it doesn't follow the [rules of hooks](https://legacy.reactjs.org/docs/hooks-rules.html), and so it produces a bunch of error messages in the console, making it so hard to debug.

image


HoonBaek (approved):
This is a good catch, thank you @angelhcp

PR #473, opened 2023-04-04
feat: replace main pkg manager npm -> yarn 3

AhyoungRyu:
### Description Of Changes
https://sendbird.atlassian.net/browse/UIKIT-3546

I highly recommend you to test this branch on your local and see if it works as before.
I checked `yarn storybook` & `yarn build` work fine on my local machine and it seems the CI passes too. 

To test this on your side, 

#### 1. Remove root `node_modules` folder
Most of deps (except storybook; it will create its own on `yarn run storybook` cmd [see why](https://github.com/storybookjs/storybook/issues/11113#issuecomment-642851848))
will be installed under `.yarn/cache` not `node_modules`
```
$ rm -r node_modules
```

#### 2. Install & enable Yarn globally
Plz refer to these steps described [here](https://yarnpkg.com/getting-started/install).
Make sure now you have Yarn 3.x.x version by printing `yarn --version` on your console.

#### Q. Do we want to enable [Zero-Installs](https://yarnpkg.com/features/zero-installs) too?
It's a powerful feature of Yarn berry for sure but it depends. Here're some pros and cons of the feature.

- Pros:
  - Faster install times, as packages are already available in the global cache <- this will reduce the dep installation time in CI
  - Reduced disk space usage, as packages are not stored locally
  - Improved consistency across different environments, as the same package versions are used globally
  - Easier management of large projects with many dependencies
  - Easier versioning of dependencies and improved reproducibility of builds
  - Everyone can use the same dep version since all deps will be managed by Git

- Cons:
  - Increased network usage, as packages need to be downloaded on demand
  - Increased risk of network-related failures, as packages may not be available in the cache or on the remote registry
  - Potentially slower startup times, as packages need to be linked at runtime


README.md
```diff
@@ -9,7 +9,7 @@ We are introducing a new version of the Sendbird Chat UIKit. Version 3 features
 
 ## Installation
 
-`npm i @sendbird/uikit-react`
```
sravan-s, line 12: Dont need to remove this, clients can use any pkg manager ~ is better to keep both :) 

`npm i @sendbird/uikit-react`
or
`yarn add @sendbird/uikit-react `

AhyoungRyu: Ahh yeah I thought about this too. Will describe both then :)

src/smart-components/App/stories/index.stories.js
```diff
@@ -1,8 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import kr from 'date-fns/locale/ko';
 
-import pkg from '../../../../package-lock.json'
```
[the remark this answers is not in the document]
sravan-s: This is a story file, `devUrl/story/app-component--version-info` . Is for QA teams/ externals to check versions. 
So, I recommend ->
`import pkg from '../../../../package.json' `

bang9:
Shouldn't we add a yarn binary and `.yarnrc.yml` file that specifies the yarn binary(.yarn/releases/*) and yarnPath?
There may be projects that still use yarn v1, so switching may be necessary.

sravan-s:
@bang9 This is for setting up dev environment, not for version that distributed through npm ~ Does it still matter?

bang9:
@sravan-s  It's an idea for the convenience of Solution Engineers who occasionally [fork](https://github.com/YJSendBird/sendbird-uikit-react) and test/work on something and maintainers.

It would be ideal if we could standardize the environment to yarn berry, but until then, frequent switching may be necessary since most repositories, including core, uikit-react-native, are currently using yarn v1.

If the yarn version is not specified within the project, the locally specified version will be used, which could lead to unexpected installations.

AhyoungRyu:
@bang9 Yeah I got your point Airen. Let me add that files too :)

bang9:
+It seems that the current installation process in CI is using yarn v1. (Adding the binary to .yarn/release seems to fix it.)
https://github.com/sendbird/sendbird-uikit-react/actions/runs/4607267036/jobs/8141571814#step:5:5

AhyoungRyu:
@bang9 Oh it was my bad. I should've set the version and commit `.yarn/` together. Fixed in https://github.com/sendbird/sendbird-uikit-react/pull/473/commits/39f2cc1db673a092a120f79b75db6a8e04a4afcf

Seems fine now https://github.com/sendbird/sendbird-uikit-react/actions/runs/4617536925/jobs/8163908954

README.md
AhyoungRyu: https://github.com/sendbird/sendbird-uikit-react/pull/473/commits/17e04b7e383044976bfdac83ff1c72248029c863 will this be okay? Feel free to re-commit if you want to touch more

AhyoungRyu:
BTW I won't merge this PR in this sprint but will do next week just to be safe.

.github/workflows/test-n-lint.yml
sravan-s, line 1: Before merging, can I ask you for couple of connivence work :D 

1. lets upgrade to `actions/checkout@v3`
2. actions/setup-node to v3
3. yarn cache -> https://github.com/actions/cache/blob/main/examples.md#node---yarn

sravan-s (approved):

.github/workflows/test-n-lint.yml
sravan-s: https://github.com/actions/cache/blob/main/examples.md#node---yarn-2

AhyoungRyu: Yeah will do separately

AhyoungRyu:
I tried to build command too. Seems there're some warnings but same happens \w npm too ;) 
Screenshot 2023-04-11 at 11 00 54 AM

sravan-s:
yeah, those are fine, I have a branch to clean that up ajjaja

AhyoungRyu:
FYI I put `npm publish` cmds back instead of updating them \w yarn in this PR since yarn's publish cmd didn't really work nicely. So we probably keep npm cmd for now until we find a better way \w yarn.

PR #479, opened 2023-04-06
fix: MessageInput - detect new lines in safari

sravan-s:
safari puts `<div>text</div>` for new lines inside content editable div(input)
Other browsers put newline or `br`

fixes: https://sendbird.atlassian.net/browse/UIKIT-3577

### DIFF
--- a/src/ui/MessageInput/index.jsx
+++ b/src/ui/MessageInput/index.jsx
@@ -329,6 +329,11 @@ const MessageInput = React.forwardRef((props, ref) => {
         } else if (node.nodeType === NodeTypes.ElementNode && node.nodeName === NodeNames.Br) {
           messageText += '\n';
           mentionTemplate += '\n';
+        } else if (node?.nodeType === NodeTypes.ElementNode && node?.nodeName === NodeNames.Div) {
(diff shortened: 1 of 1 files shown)


src/ui/MessageInput/index.jsx
```diff
@@ -329,6 +329,11 @@ const MessageInput = React.forwardRef((props, ref) => {
         } else if (node.nodeType === NodeTypes.ElementNode && node.nodeName === NodeNames.Br) {
           messageText += '\n';
           mentionTemplate += '\n';
+        } else if (node?.nodeType === NodeTypes.ElementNode && node?.nodeName === NodeNames.Div) {
```
AhyoungRyu, line 332: Q. Just curious, how do we know if it's Safari env or not? 
>node?.nodeName === NodeNames.Div

^^ This one?

sravan-s: so, div shouldnt arrive otherwise into MessageInput(checked fire fox and chrome)

Only other way dom element arrive is through paste(hope I am not wrong) -> we extract text for those cases
You think I should put strong check on browser type?

PR #481, opened 2023-04-07
feat: implement scheduler for markAsRead

sravan-s:
markAsRead is called on individual channels is unoptimal
because we have a list of channels that do this
ideally this should be fixed in server/SDK
this is a work around for the meantime to un-throttle the customer

Todo: maybe also move markAsDelivered into this format

fixes: https://sendbird.atlassian.net/browse/UIKIT-3430

### DIFF
--- a/src/lib/SendbirdState.tsx
+++ b/src/lib/SendbirdState.tsx
@@ -2,7 +2,7 @@ import type SendbirdChat from '@sendbird/chat';
 import type { User } from '@sendbird/chat';
 import type { GroupChannel, Member } from '@sendbird/chat/groupChannel';
 import { Dispatch } from 'react';
-import { MarkAsReadSchedulerType } from './hooks/useMarkAsRead';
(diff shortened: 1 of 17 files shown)


src/lib/SendbirdState.tsx
```diff
@@ -55,6 +56,7 @@ interface SendBirdStateConfig {
   };
   isTypingIndicatorEnabledOnChannelList?: boolean;
   isMessageReceiptStatusEnabledOnChannelList?: boolean;
+  markAsRead?: MarkAsReadSchedulerType;
```
HoonBaek, line 59: we need to discuss this props name

HoonBaek (commented):
LGTM, however, the props `markAsRead` doesn't look means type
markAsReadType is more meaningful (my poor idea)

+ there're lint error and conflict / need to fix before merging

HoonBaek (approved):

src/lib/SendbirdState.tsx
sravan-s: markAsReadScheduler

PR #489, opened 2023-04-12
feat: new reducer convention

sravan-s:
Implement a new convention to make reducers typesafe and maintainable
All new reducers should follow this convention and we would migrate older
reducers on the go

* Summary:
	- Specify State[Type] and InitialState in `initialState.ts`
	- Specify ACTIONS and Action[Types] in `actionTypes.ts`
		- Use helper `CreateAction`
	- Recommend using ts-pattern in reducer
See: `./src/utils/typeHelpers/reducers/README.md` for details

* Setup plop file for templating reducer
	Run: `npm run generate-component`
	Pick: `simple reducer` and specify the path where generated files should go

* CreateAction is a typescript helper to convert map into union

fixes: https://sendbird.atlassian.net/browse/UIKIT-3581

### DIFF
--- a/src/utils/typeHelpers/reducers/README.md
+++ b/src/utils/typeHelpers/reducers/README.md
@@ -5,7 +5,7 @@ and return a new state. Reducers are pure functions, which means they
 should not mutate the state, but return a new state object
 
 In this project we recommend using minimal external dependencies,
-so we recommend using built in hooks like useState and useReducer
+so we recommend using built-in hooks like useState and useReducer
 
 You can use useState if the component is simple, otherwise you can move to useReducer and move side effects into customHooks
 


src/utils/typeHelpers/reducers/README.md
```diff
@@ -0,0 +1,73 @@
+# How to define reducers
+
+Reducers are functions that take the current state and an action,
+and return a new state. Reducers are pure functions, which means they
+should not mutate the state, but return a new state object
+
+In this project we recommend using minimal external dependencies,
+so we recommend using built in hooks like useState and useReducer
```
AhyoungRyu, line 8: nit;
```suggestion
so we recommend using built-in hooks like useState and useReducer
```

src/utils/typeHelpers/reducers/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+If the component is simple, you can use useState, but if the component is
+complex, you can move to useReducer and move sideeffects into customHooks
```
AhyoungRyu, line 11: nit; how about this?

```suggestion
You can use useState if the component is simple, otherwise you can move to useReducer and move side effects into customHooks
```

src/lib/dux/sdk/thunks.ts
```diff
@@ -82,6 +107,7 @@ export const handleConnection = ({
         }
         // to check if code is released version from rollup and *not from storybook*
         // see rollup config file
+        // @ts-ignore
```
AhyoungRyu, line 110: Q. what does it say if we don't have this?

AhyoungRyu (approved):
Just left some nit comments on the documentation. LGTM 👍

PR #460, opened 2023-03-29
fix: Move scroll of message list without scrollIntoView

HoonBaek:
### Description Of Changes

Issue
* If customer had setup OpenChannel on a Page that has scroll, it used to scroll the whole page, unintentionally
* The function `scrollIntoView` has affected the outside element.
  So it triggers weird scroll behavior that we didn't expect of the parent elements.

Fix
* Move message list scroll using element.scrollTop instead of the function scrollIntoView in the OpenChannelMessageList

[UIKIT-3420](https://sendbird.atlassian.net/browse/UIKIT-3420)

[UIKIT-3420]: https://sendbird.atlassian.net/browse/UIKIT-3420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #492, opened 2023-04-14
feat: apply common scroll hook to GroupChannel MessageList

AhyoungRyu:
### Description Of Changes
Resolves [UIKIT-3620](https://sendbird.atlassian.net/browse/UIKIT-3620)

I applied the same fix logic that @HoonBaek made in https://github.com/sendbird/sendbird-uikit-react/pull/460 for OpenChannel to `smart-component/Channel/MessageList` which is in charge of GroupChannel. 
To do that, I moved useHandleOnScrollCallback hook to `src/hooks` to make reusable as I asked in [Slack](https://sendbird.slack.com/archives/C050Z6DA70A/p1681349971841599?thread_ts=1681181122.472109&cid=C050Z6DA70A). 

[UIKIT-3620]: https://sendbird.atlassian.net/browse/UIKIT-3620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/hooks/useHandleOnScrollCallback/index.ts
sravan-s, line 1: tabs here look like 4 spaces? is it intended?

AhyoungRyu: Yeah I had to rebase after merging the lint PR. Just rebased :)

AhyoungRyu:
Will ping reviewers once the storybook issue is resolved 
https://sendbird.slack.com/archives/C050Z6DA70A/p1681438203821059

HoonBaek (commented):
We need to replace the `scrollIntoView` with this custom hook `useHandleOnScrollCallback`.
Can you replace the remained those functions? There are a few places that still use scrollIntoView.

AhyoungRyu:
@HoonBaek As we said in huddle, I replaced scrollIntoView \w the hook only in Message / MessageList since I thought changing every scrollIntoView in this PR looks a bit risky. I think this change will be enough to fix [the SBIssue](https://sendbird.atlassian.net/browse/UIKIT-3620). 

Here's how it works \w this patch. The scroll shouldn't go up on entering new message like this.
after


Plz feel free to leave more comments if you have any other suggestions :)

src/hooks/useHandleOnScrollCallback/index.ts
```diff
@@ -1,25 +1,25 @@
...
 }
 
 export function useHandleOnScrollCallback({
-  setShowScrollDownButton,
   hasMore,
   onScroll,
   scrollRef,
-}: UseHandleOnScrollCallbackProps): (e: React.UIEvent<HTMLElement>) => void {
-  return useCallback((e) => {
-    const element = e.target as Element;
```
AhyoungRyu, line 22: I thought we don't need to get an event param here since we can use the scrollRef instead. I checked the behavior in OpenChannel \w Hoon to make sure I didn't break anything made by https://github.com/sendbird/sendbird-uikit-react/pull/460. 

Plus, to replace the scrollIntoView in useEffect or useLayoutEffect \w this custom hook, there was no way to get the event param. That's why I had to find other way to make things work.

src/hooks/useHandleOnScrollCallback/index.ts
```diff
@@ -1,25 +1,25 @@
 import React, { useCallback } from 'react';
-import { SCROLL_BUFFER } from '../../../../utils/consts';
+import { SCROLL_BUFFER } from '../../utils/consts';
 
 export interface UseHandleOnScrollCallbackProps {
-  setShowScrollDownButton: React.Dispatch<React.SetStateAction<boolean>>;
   hasMore: boolean;
   onScroll(fn: () => void): void;
   scrollRef: React.RefObject<HTMLDivElement>;
+  setShowScrollDownButton?: React.Dispatch<React.SetStateAction<boolean>>;
```
AhyoungRyu, line 8: Maybe there's a better way to make this logic reusable but not to break anything for now, just made this one to optional.

HoonBaek (approved):
LGTM

PR #497, opened 2023-04-17
fix: Bring back version of deps. to storybook

sravan-s:
When we removed package-lock, this info was removed~


AhyoungRyu (approved):
Oh thanks Sravan 😄

HoonBaek (approved):

PR #498, opened 2023-04-18
feat: Convert connection-thunk to useConnect hook

sravan-s:
Move logic from sdk/thunks to hooks/useConnect
```
  const reconnect = useConnect({
    appId,
    userId,
    accessToken,
  }, {
    logger,
    nickname,
    profileUrl,
    configureSession,
    customApiHost,
    customWebSocketHost,
    sdk: sdkStore?.sdk,
    sdkDispatcher,
    userDispatcher,
  });
```
Add unit tests
fixes: https://sendbird.atlassian.net/browse/UIKIT-3581


src/lib/hooks/useConnect/connect.ts
```diff
@@ -0,0 +1,38 @@
...
+  accessToken,
+  sdk,
+}: ConnectTypes): Promise<void> {
+  await disconnectSdk({
+    logger,
+    sdkDispatcher,
+    userDispatcher,
+    sdk,
+  });
+  await setUpConnection({
```
AhyoungRyu, line 25: Should we wrap here \w try - catch just in case disconnectSdk or setUpConnection throws an error?

HoonBaek (approved):
LGTM

src/lib/hooks/useConnect/connect.ts
sravan-s: Before, we used to throw it to customer, can consider wrapping, yeah

sravan-s: Ah, actually wrapped in `useConnect` ~

PR #499, opened 2023-04-18
feature: Create new UI component Toggle

HoonBaek:
### Description Of Changes

Dev Docs: https://sendbird.atlassian.net/wiki/spaces/UK/pages/2061664531/Toggle+component

path
```javascript
import { Toggle, ToggleContainer, ToggleUI,  useToggleContext } from '@sendbird/ui/Toggle';
```

Provide
* ToggleContainer: A context provider component that has only logic of managing the toggle status
* ToggleUI: An UI component that doesn't include managing logic
* Toggle: A combination of ToggleContainer and ToggleUI components
* useToggleContext: A custom `useContext` hook that you can get the context from ToggleContainer

[UIKIT-3631](https://sendbird.atlassian.net/browse/UIKIT-3631)

[UIKIT-3631]: https://sendbird.atlassian.net/browse/UIKIT-3631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -1,5 +1,3 @@
-import { ChangeEventHandler } from 'react';
-
 /**
  * Type Definitions for @sendbird/uikit-react@{{ version }}
  * homepage: https://sendbird.com/
@@ -2478,9 +2476,9 @@ declare module '@sendbird/uikit-react/ui/Toggle' {
     defaultChecked?: boolean;
     disabled?: boolean;
     reversed?: boolean;
-    onChange?: ChangeEventHandler<HTMLInputElement>;
-    onFocus?: ChangeEventHandler<HTMLInputElement>;
-    onBlur?: ChangeEventHandler<HTMLInputElement>;
+    onChange?: React.ChangeEventHandler<HTMLInputElement>;
+    onFocus?: React.ChangeEventHandler<HTMLInputElement>;
+    onBlur?: React.ChangeEventHandler<HTMLInputElement>;
   }
   export interface ToggleContainerProps extends ToggleContextInterface {
     children?: React.ReactElement;
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -2475,7 +2475,6 @@ declare module '@sendbird/uikit-react/ui/Toggle' {
     checked?: boolean;
     defaultChecked?: boolean;
     disabled?: boolean;
-    reversed?: boolean;
     onChange?: React.ChangeEventHandler<HTMLInputElement>;
     onFocus?: React.ChangeEventHandler<HTMLInputElement>;
     onBlur?: React.ChangeEventHandler<HTMLInputElement>;
@@ -2484,6 +2483,7 @@ declare module '@sendbird/uikit-react/ui/Toggle' {
     children?: React.ReactElement;
   }
   export interface ToggleUIProps {
+    reversed?: boolean;
     width?: string;
     animationDuration?: string;
     style?: Record<string, string>;
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -1,3 +1,5 @@
+import { ChangeEventHandler } from 'react';
+
 /**
  * Type Definitions for @sendbird/uikit-react@{{ version }}
  * homepage: https://sendbird.com/
@@ -2468,6 +2470,37 @@ declare module '@sendbird/uikit-react/ui/ThumbnailMessageItemBody' {
   }
   const ThumbnailMessageItemBody: React.FC<ThumbnailMessageItemBodyProps>;
   export default ThumbnailMessageItemBody;
+}
+
+declare module '@sendbird/uikit-react/ui/Toggle' {
+  export interface ToggleContextInterface {
+    checked?: boolean;
+    defaultChecked?: boolean;
+    disabled?: boolean;
+    reversed?: boolean;
+    onChange?: ChangeEventHandler<HTMLInputElement>;
+    onFocus?: ChangeEventHandler<HTMLInputElement>;
+    onBlur?: ChangeEventHandler<HTMLInputElement>;
+  }
+  export interface ToggleContainerProps extends ToggleContextInterface {
+    children?: React.ReactElement;
+  }
+  export interface ToggleUIProps {
+    className?: string;
+    width?: string;
+    animationDuration?: string;
+    style?: Record<string, string>;
+    name?: string;
+    id?: string;
+    ariaLabel?: string;
+    ariaLabelledby?: string;
+  }
+  export interface ToggleProps extends ToggleContainerProps, ToggleUIProps {}
+  
+  export const Toggle: React.FC<ToggleProps>;
+  export const ToggleContainer: React.FC<ToggleContainerProps>;
+  export const ToggleUI: React.FC<ToggleUIProps>;
+  export function useToggleContext(): ToggleContextInterface;
 }
 
 declare module '@sendbird/uikit-react/ui/Tooltip' {
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -2486,7 +2486,6 @@ declare module '@sendbird/uikit-react/ui/Toggle' {
     children?: React.ReactElement;
   }
   export interface ToggleUIProps {
-    className?: string;
     width?: string;
     animationDuration?: string;
     style?: Record<string, string>;
@@ -2495,8 +2494,9 @@ declare module '@sendbird/uikit-react/ui/Toggle' {
     ariaLabel?: string;
     ariaLabelledby?: string;
   }
-  export interface ToggleProps extends ToggleContainerProps, ToggleUIProps {}
-  
+  export interface ToggleProps extends ToggleContainerProps, ToggleUIProps {
+    className?: string;
+  }
   export const Toggle: React.FC<ToggleProps>;
   export const ToggleContainer: React.FC<ToggleContainerProps>;
   export const ToggleUI: React.FC<ToggleUIProps>;
--- a/src/ui/Toggle/ToggleContainer.tsx
+++ b/src/ui/Toggle/ToggleContainer.tsx
@@ -15,7 +15,6 @@ export function ToggleContainer({
   checked = Dvalue.checked, // null
   defaultChecked = Dvalue.defaultChecked,
   disabled = Dvalue.disabled,
-  reversed = Dvalue.reversed,
   onChange = Dvalue.onChange,
   onFocus = Dvalue.onFocus,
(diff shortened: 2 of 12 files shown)


scripts/index_d_ts
```diff
@@ -1,3 +1,5 @@
+import { ChangeEventHandler } from 'react';
```
HoonBaek, line 1: Is this okay?

HoonBaek:
Uploading Record of Toggle.mp4…

HoonBaek:
https://user-images.githubusercontent.com/46333979/232713541-fbe203e2-b886-4c98-aa83-17dea32caa70.mp4

scripts/index_d_ts
sravan-s: probably better move inside `SendbirdUIKitGlobal`,

sravan-s: actually, see comment below

scripts/index_d_ts
```diff
@@ -2470,6 +2472,37 @@ declare module '@sendbird/uikit-react/ui/ThumbnailMessageItemBody' {
   export default ThumbnailMessageItemBody;
 }
 
+declare module '@sendbird/uikit-react/ui/Toggle' {
+  export interface ToggleContextInterface {
+    checked?: boolean;
+    defaultChecked?: boolean;
+    disabled?: boolean;
+    reversed?: boolean;
+    onChange?: ChangeEventHandler<HTMLInputElement>;
```
sravan-s, line 2481: React.ChangeEventHandler<HTMLInputElement> doesnt work?

sravan-s (commented):
Nice job, let me check

scripts/index_d_ts
HoonBaek: Ah yes, that's good

src/ui/Toggle/ToggleContext.ts
```diff
@@ -0,0 +1,34 @@
...
+  reversed?: boolean;
+  onChange?: ChangeEventHandler<HTMLInputElement>;
+  onFocus?: ChangeEventHandler<HTMLInputElement>;
+  onBlur?: ChangeEventHandler<HTMLInputElement>;
+}
+
+export const ToggleContext = React.createContext<ToggleContextInterface>(TOGGLE_DEFAULT_VALUE);
+
+export function useToggleContext(): ToggleContextInterface {
+  const context = React.useContext(ToggleContext);
```
sravan-s, line 32: good to check & throw error when context is empty ~

src/ui/Toggle/ToggleUI.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const toggleWidth = useMemo(() => (
+    filterNumber(width)?.[0]
+  ), [width]);
+  const toggleHeight = useMemo(() => (
+    toggleWidth / 2
+  ), [toggleWidth]);
+  const dotSize = useMemo(() => (
+    toggleHeight * 0.6
+    // The size of dot should be 60% of toggle height
+  ), [toggleHeight]);
```
sravan-s, line 46: Dont need to memoize these :) just use normal number

src/ui/Toggle/index.scss
```diff
@@ -0,0 +1,82 @@
+@import '../../styles/variables';
+
+.sendbird-input-toggle-button {
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+
+  box-shadow: 2px solid red;
```
sravan-s, line 8: box-shadow: red >  same for both themes? :curious:

HoonBaek: it's a legacy, let me remove it thanks

src/ui/Toggle/ToggleContext.ts
HoonBaek: Could you show me an example?

sravan-s:
```
  const context = React.useContext(MessageContext);
  if (context === undefined) {
    throw new Error('useMessageContext must be used within a MessageProvider');
  }
  return context;
```

src/ui/Toggle/__tests__/__snapshots__/Toggle.spec.js.snap
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Toggle should do a snapshot test of the Toggle DOM 1`] = `
+<DocumentFragment>
+  <div
+    class="sendbird-ui-toggle undefined"
+  >
+    <label
+      class="sendbird-input-toggle-button  --unchecked"
```
sravan-s, line 9: To better to follow BEM (atleast as we are trying to), for consistency, the classnames should be `sendbird-input-toggle-button  sendbird-input-toggle-button--unchecked `

src/ui/Toggle/__tests__/__snapshots__/Toggle.spec.js.snap
```diff
@@ -0,0 +1,26 @@
...
+      <div
+        class="sendbird-input-toggle-button__inner-dot sendbird-input-toggle-button__inner-dot--inactivate"
+        style="width: 12px; height: 12px; animation-duration: 0.5s;"
+      />
+      <input
+        aria-label=""
+        aria-labelledby=""
+        id=""
+        name=""
+        type="checkbox"
```
sravan-s, line 21: good to have one more snapshot with these values on

src/ui/Toggle/index.scss
```diff
@@ -0,0 +1,81 @@
...
+  position: absolute;
+  border-radius: 50%;
+  @include themed() {
+    background-color: t(bg-0);
+  }
+  animation-iteration-count: 1;
+}
+
+/* Manage animation and position by status */
+@keyframes moveToRight {
```
sravan-s, line 46: maybe call these `sendbirdMoveToRight` ~

src/ui/Toggle/index.scss
```diff
@@ -0,0 +1,81 @@
...
+  0% { right: 60%; }
+  100% { right: 10%; }
+}
+@keyframes moveToLeft {
+  0% { right: 10%; }
+  100% { right: 60%; }
+}
+// normal - animation
+.sendbird-input-toggle-button.--turned-on .sendbird-input-toggle-button__inner-dot {
```
sravan-s, line 55: from previous comment about BEM:  `.sendbird-input-toggle-button--turned-on`

src/ui/Toggle/ToggleContext.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const ToggleContext = React.createContext<ToggleContextInterface>(TOGGLE_DEFAULT_VALUE);
+
+export function useToggleContext(): ToggleContextInterface {
+  let context = null;
+  try {
+    context = React.useContext(ToggleContext);
+    return context;
+  } catch (_) { }
+  return context;
```
sravan-s, line 35: Ah.. from previous comment, I wanted to say, its okay to throw error here:
It would be caught at development side while developer is implementing :) 

```
  const context = React.useContext(MessageContext);
  if (context === undefined) {
    throw new Error('useToggleContext must be used within a ToggleContainer');
  }
  return context;
```

HoonBaek: Ah, yeah I felt I understood wrong. Thanks

src/ui/Toggle/__tests__/__snapshots__/Toggle.spec.js.snap
HoonBaek: I was worried about the query selector became too longer in CSS

src/ui/Toggle/ToggleContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+  checked = Dvalue.checked, // null
+  defaultChecked = Dvalue.defaultChecked,
+  disabled = Dvalue.disabled,
+  onChange = Dvalue.onChange,
+  onFocus = Dvalue.onFocus,
+  onBlur = Dvalue.onBlur,
+  children = null,
+}: ToggleContainerProps): React.ReactElement {
+  const [isChecked, setChecked] = useState(defaultChecked || false);
+  const useOnChangeCallback = useCallback((e: ChangeEvent<HTMLInputElement>) => {
```
AhyoungRyu, line 24: nit; 
```suggestion
  const handleChange = useCallback((e: ChangeEvent<HTMLInputElement>) => {
```
this kind of shorten name \wo `use` prefix would be fine here

src/ui/Toggle/utils.ts
```diff
@@ -0,0 +1,10 @@
+export function filterNumber(str: string): Array<number> {
+  if (typeof str === 'number') {
+    return [str];
+  } else if (typeof str !== 'string') {
+    return [];
+  }
```
AhyoungRyu, line 6: Since the param type is 'string', this case won't be exist I guess. 
```
filterNumber(111) 
// or
filterNumber({...})
```
Both will throw an type error from beginning.

src/ui/Toggle/ToggleUI.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const {
+    checked,
+    disabled,
+    onChange,
+    onFocus,
+    onBlur,
+  } = useToggleContext();
+
+  // animation should not be activated in the initialization step
+  const [animatedClassName, setAnimated] = useState('');
```
AhyoungRyu, line 36:
```suggestion
  const [animatedClassName, setAnimatedClassName] = useState('');
```
Why don't we give the exact name for the dispatcher?

AhyoungRyu (approved):
Great work! Just left some nit comments which are not blockers.

src/ui/Toggle/utils.ts
HoonBaek: I think the error will be shown for TypeScript, but we are supporting pure JavaScript too.
anyway, I applied for the review, could you review it again? @AhyoungRyu

AhyoungRyu: Ahh I see. Thank you for the update :)

AhyoungRyu (approved):

src/ui/Toggle/index.scss
```diff
@@ -0,0 +1,81 @@
+@import '../../styles/variables';
+
+.sendbird-input-toggle-button {
```
sravan-s, line 3: have recommendation to rewrite this file, will explain

HoonBaek: I applied that!

sravan-s (approved):

PR #501, opened 2023-04-21
fix: Use dynamic import for VoiceRecorder

HoonBaek:
### Description Of Changes

* Use dynamic import to lazy load the audio processor only when using the voice message feature

[UIKIT-3590](https://sendbird.atlassian.net/browse/UIKIT-3590)

[UIKIT-3590]: https://sendbird.atlassian.net/browse/UIKIT-3590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/VoiceRecorder/index.tsx
+++ b/src/hooks/VoiceRecorder/index.tsx
@@ -1,5 +1,4 @@
-import React, { createContext, useCallback, useContext, useState } from 'react';
-import { downsampleToWav, encodeMp3 } from './WebAudioUtils';
+import React, { createContext, useCallback, useContext, useEffect, useState } from 'react';
 import {
   VOICE_MESSAGE_FILE_NAME,
   VOICE_MESSAGE_MIME_TYPE,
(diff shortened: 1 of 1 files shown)


src/hooks/VoiceRecorder/index.tsx
```diff
@@ -34,11 +33,25 @@ const VoiceRecorderContext = createContext<VoiceRecorderContext>({
...
   const [isRecordable, setIsRecordable] = useState<boolean>(false);
 
+  const [webAudioUtils, setWebAudioUtils] = useState(null);
+  useEffect(() => {
+    if (isVoiceMessageEnabled && !webAudioUtils) {
+      import('./WebAudioUtils').then((data) => {
+        setWebAudioUtils(data);
+      })
+    }
+  }, []);
```
AhyoungRyu, line 47: Maybe we can have a separate hook like 

```
const useLazyLoading({
  cond,
  module,
}) {
  const [data, setData] = useState(null);
  useEffect(() => {
    if (cond) {
      import(module).then((data) => {
        setData(data);
      })
    }
  }, []);
  
  return null;
}
```

and its usage would be like 
```
useLazyLoading({ cond: ..., module: ... });
```

AhyoungRyu (approved):

src/hooks/VoiceRecorder/index.tsx
HoonBaek: Yeah, that's a good idea, let me keep this topic to the next sprint

src/hooks/VoiceRecorder/index.tsx
```diff
@@ -34,11 +33,25 @@ const VoiceRecorderContext = createContext<VoiceRecorderContext>({
 export const VoiceRecorderProvider = (props: VoiceRecorderProps): React.ReactElement => {
   const { children } = props;
   const { config } = useSendbirdStateContext();
-  const { logger } = config;
+  const { logger, isVoiceMessageEnabled } = config;
   const [mediaRecorder, setMediaRecorder] = useState<MediaRecorder>(null);
   const [isRecordable, setIsRecordable] = useState<boolean>(false);
 
+  const [webAudioUtils, setWebAudioUtils] = useState(null);
```
sravan-s, line 40: Hmm, here, maybe it was better to have a simple audioUtil, that doesn't convert file to mp3

sravan-s: cc @HoonBaek

HoonBaek:
image

PR #507, opened 2023-04-26
chore: replace nodeLinker node_modules \w pnpm

AhyoungRyu:
### Description Of Changes
More details about this change are described in [UIKIT-3771](https://sendbird.atlassian.net/browse/UIKIT-3771) :) 

[UIKIT-3771]: https://sendbird.atlassian.net/browse/UIKIT-3771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
     "stylelint": "^13.0.0",
     "stylelint-config-sass-guidelines": "^7.0.0",
     "ts-loader": "^7.0.5",
-    "ts-pattern": "^4.2.2",
+    "ts-pattern": "4.0.5",
     "tslib": "^2.0.0",
     "typescript": "^4.7.4"
   },
--- a/src/lib/dux/sdk/reducers.ts
+++ b/src/lib/dux/sdk/reducers.ts
@@ -4,14 +4,14 @@ import initialState, { SdkStoreStateType } from './initialState';
 
 export default function reducer(state: SdkStoreStateType, action: SdkActionTypes): SdkStoreStateType {
   return match(action)
-    .with({ type: SDK_ACTIONS.SET_SDK_LOADING }, ({ payload }) => {
+    .with(SDK_ACTIONS.SET_SDK_LOADING, ({ payload }) => {
       return {
         ...state,
         initialized: false,
         loading: payload,
       };
     })
-    .with({ type: SDK_ACTIONS.SDK_ERROR }, () => {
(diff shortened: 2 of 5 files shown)


src/lib/dux/sdk/reducers.ts
```diff
@@ -4,30 +4,30 @@ import initialState, { SdkStoreStateType } from './initialState';
 
 export default function reducer(state: SdkStoreStateType, action: SdkActionTypes): SdkStoreStateType {
   return match(action)
```
sravan-s, line 6: should be action.type

src/lib/dux/sdk/reducers.ts
```diff
@@ -4,30 +4,30 @@ import initialState, { SdkStoreStateType } from './initialState';
...
-    .with({ type: SDK_ACTIONS.SET_SDK_LOADING }, ({ payload }) => {
+    .with(SDK_ACTIONS.SET_SDK_LOADING, ({ payload }) => {
```
sravan-s, line 7: So, if we match(action.type), we would still get `({ payload })` ? It didnt work that way sometime before I checked

AhyoungRyu: ahh yeah my bad. let me update again

package.json
```diff
@@ -140,7 +140,7 @@
     "stylelint": "^13.0.0",
     "stylelint-config-sass-guidelines": "^7.0.0",
     "ts-loader": "^7.0.5",
-    "ts-pattern": "^4.2.2",
+    "ts-pattern": "4.0.5",
```
sravan-s, line 143: This fixed it?

AhyoungRyu: Nahh. I could get more understandable error with this version(4.0.5) since ts-pattern's version and ours were nor compatible. But still getting the error unfortunately. 

Now I'm thinking 
`nodeLinker: node-modules <-> pnpm (or pnp)` impacted by rollup config. Probably we just need to use node-modules 😅

AhyoungRyu: Let me check more when I have spare time.

sravan-s: yeah, is not big priority

AhyoungRyu:
Will open again once it's ready

PR #509, opened 2023-04-28
fix: modify url regex to filter

AhyoungRyu:
### Description Of Changes
Resolves [UIKIT-3707](https://sendbird.atlassian.net/browse/UIKIT-3707)

**AS-IS**
A string \w $ sign(e.g. `$123.12`) or @ sign(e.g.`team_engineering@inkle.io`) created unnecessary underline with link
before

**TO-BE**

after

[UIKIT-3707]: https://sendbird.atlassian.net/browse/UIKIT-3707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/__tests__/utils.spec.js
+++ b/src/utils/__tests__/utils.spec.js
@@ -1,4 +1,4 @@
-import { binarySearch } from '../index';
+import { binarySearch, isUrl } from '../index';
 
 describe('Global-utils', () => {
   it('should find right index with binarySearch', () => {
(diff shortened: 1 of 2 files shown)


HoonBaek (approved):
Thanks for improving the RegEx and adding tests.

What do you think about using `fetch API` to validate? @sravan-s @AhyoungRyu 
It will take more time to validate but we won't have to consider this anymore

sravan-s (approved):

AhyoungRyu:
@HoonBaek Yeah sounds like a idea too!

sravan-s:
@HoonBaek , actually, share an example :)

PR #514, opened 2023-05-02
hotfix: Separate exporting components and types

HoonBaek:
Fix the build error issue 
> [!] Error: 'ToggleContainerProps' is not exported by src/ui/Toggle/ToggleContainer.tsx, imported by src/ui/Toggle/index.tsx

### DIFF
--- a/src/ui/Toggle/ToggleContainer.tsx
+++ b/src/ui/Toggle/ToggleContainer.tsx
@@ -7,7 +7,7 @@ import {
 } from './ToggleContext';
 
 export interface ToggleContainerProps extends ToggleContextInterface {
-  children?: React.ReactElement;
+  children?: React.ReactElement | null;
 }
 
 // Props Explanation https://github.com/aaronshaf/react-toggle#props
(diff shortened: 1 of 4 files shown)


src/ui/Toggle/ToggleContainer.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from './ToggleContext';
 
 export interface ToggleContainerProps extends ToggleContextInterface {
-  children?: React.ReactElement;
+  children?: React.ReactElement | null;
```
sravan-s, line 10: children can be null?

HoonBaek: yeah, it should be nullable because the default value is null

sravan-s: But, toggle container should have childeren, no?
```
<ToggleContainer>
   ... come UI component
</ToggleContainer>
```

`<ToggleContainer/>` is useless, no?

sravan-s (approved):

PR #518, opened 2023-05-04
fix: Add types of GroupChat and OpenChat for SendbirdChat

HoonBaek:
### Description Of Changes

Issue
* Customer had to use the type assertion for removing type error
```javascript
const globalState = useSendbirdStateContext()
const getSdk = sendbirdSelectors.getSdk(globalState) as SendbirdGroupChat;
getSdk.groupChannel.getTotalUnreadMessageCount();
```

Fix
* Specify the `getSdk` returns SendbirdGroupChat or SendbirdOpenChat

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -5,7 +5,6 @@
  */
 declare module "SendbirdUIKitGlobal" {
   import type React from 'react';
-  import type Locale from 'date-fns';
   import type SendbirdChat from '@sendbird/chat';
   import type {
     SendbirdError,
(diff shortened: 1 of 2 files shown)


AhyoungRyu (approved):

bang9:
@HoonBaek 
FYI, the type of module added to the SDK can be defined as follows!

```ts
import type { ModuleNamespaces } from '@sendbird/chat/lib/__definition';
import type SendbirdChat from '@sendbird/chat';
import type { GroupChannelModule } from '@sendbird/chat/groupChannel';
import type { OpenChannel } from '@sendbird/chat/openChannel';

export type SendbirdChatSDK = SendbirdChat & ModuleNamespaces<[GroupChannelModule, OpenChannelModule]>;
```

HoonBaek:
@bang9 Thanks for good suggestion, I'm curious
```
ModuleNamespaces<[GroupChannelModule, OpenChannelModule]>
```
is it okay in the case that the SendbirdChat contains only one of the modules

I guessed this is more preferred
```
ModuleNamespaces<[GroupChannelModule | OpenChannelModule]>
```

I couldn't be wrong, so point it out, I really appreciate

bang9:
@HoonBaek The module namespace type accepts an **array of modules** as generics.
If an SDK instance should only contain one module, it is correct to receive only one.
If two SDK instances each have only one module, one for open channel and one for group channel, the types should also be separated. (Typically, types should follow the actual code.)

PR #526, opened 2023-05-09
chore: update eslint & its plugin version

AhyoungRyu:
### Description Of Changes
Aims to resolve [UIKIT-1715](https://sendbird.atlassian.net/browse/UIKIT-1715) and [UIKIT-3841](https://sendbird.atlassian.net/browse/UIKIT-3841).

Updated ESLint version to latest and related plugins too. 
It seems some of rules have been turned on by default in the latest plugins. So I had to modify some code or disable some accordingly. 

With this change, these below error and warning should be gone too.
Screenshot 2023-05-09 at 3 22 18 PM

[UIKIT-1715]: https://sendbird.atlassian.net/browse/UIKIT-1715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UIKIT-3841]: https://sendbird.atlassian.net/browse/UIKIT-3841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.eslintignore
```diff
@@ -3,7 +3,6 @@ __test__/*/**.js
 __test__/*/**.jsx
 *.spec.js
 *.mock.js
-src/ui/MessageInput/index.jsx
```
AhyoungRyu, line 6: Now it's okay with not having this here.

.eslintrc.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     Atomics: 'readonly',
     SharedArrayBuffer: 'readonly',
   },
-  parser: 'babel-eslint',
```
AhyoungRyu, line 15: The name of parser has been changed in the new version of eslint.

src/index.ts
AhyoungRyu, line 1: I changed the format of `src/index` file from `js` to `ts` to resolve this weird warning thrown by eslint-ts. 
Screenshot 2023-05-09 at 3 28 39 PM

src/modules/Channel/context/utils.js
```diff
@@ -8,6 +8,7 @@ import { OutgoingMessageStates } from '../../../utils/exports/getOutgoingMessage
 const UNDEFINED = 'undefined';
 const { SUCCEEDED, FAILED, PENDING } = getSendingMessageStatus();
 
+/* eslint-disable default-param-last */
```
AhyoungRyu, line 11: Disabled the rule here not to make breaking change like changing param order.

src/ui/VoiceMessageInput/types.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-redeclare */
```
AhyoungRyu, line 1: It's because we have duplicated declarations here like
```
export const VoiceMessageInputStatus = 
export type VoiceMessageInputStatus =
```

There are several places we're doing like this. I disabled the rule for them for now but I guess we'd better give different name for variable & type next time. It's a bit hard to differenciate  which one is variable or type.

sravan-s: I wanted different name for this ~ 
`VoiceMessageInputStatus` & `VoiceMessageInputStatusType` cc @HoonBaek

.eslintrc.json
```diff
@@ -63,6 +73,7 @@
     "no-bitwise": "off",
     "no-restricted-syntax": "off",
     "no-prototype-builtins": "off",
+    "no-unsafe-optional-chaining": "off",
```
sravan-s, line 76: isnt it better to have this?

.eslintrc.json
```diff
@@ -35,6 +44,7 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-unused-expressions": "error",
     "react/display-name": "off",
+    "react/function-component-definition": "off",
```
sravan-s, line 47: this only removes type defn from react components, right?
 normal functions:
```
function concat(a: numb, b: num): string { return `${a}${b}` }
```
will have return type defn, correct?

bundle-analysis.json
sravan-s, line 1: dont need to commit this ~

sravan-s (approved):

src/ui/VoiceMessageInput/types.ts
HoonBaek: Ah this is my legacy..

AhyoungRyu: Oh then better to give them different names and remove `eslint-disable` too later.

bundle-analysis.json
AhyoungRyu: Ahh yeah I dropped the change https://github.com/sendbird/sendbird-uikit-react/pull/526/commits/c60445827f31e38c5869e6b236dc85afcec02380. It's changed when I tried to see if the build is okay. 

BTW do u think we'd better add `bundle-analysis.json` to .gitignore list? Guess the file is only used when we create a production build.

sravan-s: I was thinking about removing it altogether, maybe keep it with testing part of nightly ~

PR #527, opened 2023-05-09
chore: turn on no-unsafe-optional-chaining rule again

AhyoungRyu:
### Description Of Changes
Addressed a comment in https://github.com/sendbird/sendbird-uikit-react/pull/526#discussion_r1188223665 to bring back `no-unsafe-optional-chaining` rule again which is useful to have. 

As I mentioned [here](https://sendbird.slack.com/archives/C050Z6DA70A/p1683624307711019),  I found there're several places(actually many) we're doing optional chaining unnecessarily. Some of them are caught by lint(fixed in this PR) but others are still there. We'd better revisit them too.

### DIFF
--- a/src/hooks/VoicePlayer/useVoicePlayer.tsx
+++ b/src/hooks/VoicePlayer/useVoicePlayer.tsx
@@ -70,9 +70,9 @@ export const useVoicePlayer = ({
     play: playVoicePlayer,
     pause: pauseVoicePlayer,
     stop: stopVoicePlayer,
-    playbackTime: currentAudioUnit?.playbackTime * 1000,
-    duration: currentAudioUnit?.duration * 1000,
+    playbackTime: currentAudioUnit.playbackTime * 1000,
+    duration: currentAudioUnit.duration * 1000,
     // the unit of playbackTime and duration should be millisecond
-    playingStatus: currentAudioUnit?.playingStatus,
+    playingStatus: currentAudioUnit.playingStatus,
   });
 };
--- a/src/modules/Thread/components/ParentMessageInfo/index.tsx
+++ b/src/modules/Thread/components/ParentMessageInfo/index.tsx
@@ -42,7 +42,7 @@ export default function ParentMessageInfo({
     isOnline,
     userMention,
   } = config;
-  const userId = stores.userStore.user.userId;
+  const userId = stores.userStore.user?.userId ?? '';
   const { dateLocale } = useLocalization();
   const {
     currentChannel,
--- a/src/modules/Thread/components/ParentMessageInfo/index.tsx
+++ b/src/modules/Thread/components/ParentMessageInfo/index.tsx
@@ -42,8 +42,8 @@ export default function ParentMessageInfo({
     isOnline,
     userMention,
   } = config;
-  const userId = stores?.userStore?.user?.userId;
-  const { dateLocale } = useLocalization?.();
+  const userId = stores.userStore.user.userId;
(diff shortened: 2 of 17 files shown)


src/hooks/VoicePlayer/useVoicePlayer.tsx
```diff
@@ -70,9 +70,9 @@ export const useVoicePlayer = ({
     play: playVoicePlayer,
     pause: pauseVoicePlayer,
     stop: stopVoicePlayer,
-    playbackTime: currentAudioUnit?.playbackTime * 1000,
```
AhyoungRyu, line 73: Based on `currentAudioUnit` type, don't need the optional chaining here.

src/modules/Thread/components/ParentMessageInfo/index.tsx
```diff
@@ -42,8 +42,8 @@ export default function ParentMessageInfo({
     isOnline,
     userMention,
   } = config;
-  const userId = stores?.userStore?.user?.userId;
```
AhyoungRyu, line 45: Seems every objects' type here are not nullable. So removed all `?` but plz tell me if we need to be safe here. Then I'll correct each types as nullable :)

sravan-s: Removing all `?` is okay ~

src/ui/MessageContent/index.tsx
```diff
@@ -266,7 +266,7 @@ export default function MessageContent({
               message={message as UserMessage | FileMessage}
               userId={userId}
               isByMe={isByMe}
-              isUnavailable={(replyType === 'THREAD' && (channel?.joinedAt * 1000) > message?.parentMessage?.createdAt)}
+              isUnavailable={(replyType === 'THREAD' && (channel.joinedAt * 1000) > (message.parentMessage?.createdAt ?? 0))}
```
sravan-s, line 269: `message.parentMessage ` cc @HoonBaek

src/utils/index.ts
```diff
@@ -163,11 +163,11 @@ export const isVoiceMessage = (message: FileMessage): boolean => {
   if (!(message && isFileMessage(message))) {
     return false;
   }
-  const [mimeType, typeParameter] = message.type?.split(';');
+  const [mimeType, typeParameter] = message.type.split(';');
```
sravan-s, line 166: cc @HoonBaek

sravan-s (approved):
Wait for @HoonBaek

src/ui/MessageContent/index.tsx
AhyoungRyu: @HoonBaek  Its type looked okay without the optional chaining, but let me know if there's any concern :)

src/utils/index.ts
HoonBaek: This looks fine

src/modules/Thread/components/ParentMessageInfo/index.tsx
HoonBaek: userId could be undefined
so how about
```
const userId = stores.userStore.user?.userId;
```

src/ui/MessageContent/index.tsx
HoonBaek: I'm worried about the case when the channel is null, how can we manage the case

HoonBaek (approved):
I added the Nullable type and operated the case when the channel is empty

AhyoungRyu:
@HoonBaek Thanks!

PR #530, opened 2023-05-11
feature: Thread for mobile screen

HoonBaek:
### Description Of Changes

* Apply MobileMenu/BottomSheet to the components: ParentMessageInfo and ThreadListItem
* Apply the Thread module to the MobileLayout component
* Add new Icon Thread

[UIKIT-3476](https://sendbird.atlassian.net/browse/UIKIT-3476)

[UIKIT-3476]: https://sendbird.atlassian.net/browse/UIKIT-3476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Thread/components/ParentMessageInfo/index.tsx
```diff
@@ -192,7 +208,11 @@ export default function ParentMessageInfo({
   }
 
   return (
-    <div className={`sendbird-parent-message-info ${className}`}>
+    <div
+      className={`sendbird-parent-message-info ${className}`}
+      {...(isMobile) ? { ...longPress } : {}}
```
AhyoungRyu, line 213:
```suggestion
      {...isMobile ? longPress : {}}
```
I guess you can get the same result \w the above :)

src/modules/Thread/components/ThreadList/ThreadListItemContent.tsx
```diff
@@ -88,8 +92,23 @@ export default function ThreadListItemContent({
...
+        setShowMobileMenu(true);
+      }
+    },
+  });
+
   return (
-    <div className={`sendbird-thread-list-item-content ${className} ${isByMe ? 'outgoing' : 'incoming'}`}>
+    <div
+      className={`sendbird-thread-list-item-content ${className} ${isByMe ? 'outgoing' : 'incoming'}`}
+      {...(isMobile) ? { ...longPress } : {}}
```
AhyoungRyu, line 109:
```suggestion
      {...isMobile ? longPress : {}}
```
ditto

src/lib/MediaQueryContext.tsx
```diff
@@ -11,14 +11,14 @@ const MediaQueryContext = React.createContext({
 
 export interface MediaQueryProviderProps {
   children: React.ReactElement;
-  mediaQueryBreakPoint?: string | boolean;
+  mediaQueryBreakPoint?: string;
```
sravan-s, line 14: `string | boolean ` is a conscious choice.. If user wants to target a specific condition externally, they can make it true(say they dont want mobile mode on certain tablets, even when size is below certain pixels).. I was thinking its more simple than having 2 separate props

sravan-s: How about we call it `responsive` ?

`responsive?: string | boolean;`

src/modules/App/MobileLayout.tsx
```diff
@@ -37,18 +39,21 @@ export const MobileLayout: React.FC<MobileLayoutProps> = (
...
   const [panel, setPanel] = useState(PANELS?.CHANNEL_LIST);
+  const [animatedMessageId, setAnimatedMessageId] = useState<number | null>(null);
 
   const store = useSendbirdStateContext();
   const sdk = store?.stores?.sdkStore?.sdk as SendbirdGroupChat;
   const userId = store?.config?.userId;
 
-  const goToMessage = (message?: BaseMessage | null) => {
-    setStartingPoint(message?.createdAt);
+  const goToMessage = (message?: BaseMessage | null, timeoutCb?: (msgId: number | null) => void) => {
```
sravan-s, line 52: 👍

src/lib/MediaQueryContext.tsx
sravan-s: `breakpoint` is also nice imo

src/modules/App/DesktopLayout.tsx
```diff
@@ -44,8 +44,8 @@ export const DesktopLayout: React.FC<DesktopLayoutProps> = (
           onProfileEditSuccess={onProfileEditSuccess}
           disableAutoSelect={disableAutoSelect}
           onChannelSelect={(channel) => {
-            setStartingPoint(null);
-            setHighlightedMessage(null);
+            setStartingPoint?.(null);
+            setHighlightedMessage?.(null);
```
sravan-s, line 48: Cleander would be to make these two fns. not nullable
ie: `setStartingPoint(null); ` instead of `setStartingPoint?.(null); `

sravan-s: similar for these setters than can be shared

sravan-s: https://github.com/sendbird/sendbird-uikit-react/blob/feature/UIKIT-3476/Mobile-support-for-thread/src/modules/App/types.ts#LL30C3-L30C19

src/modules/App/mobile.scss
```diff
@@ -6,3 +6,14 @@
...
+.sb_mobile__panelwrap .sendbird-thread {
+  width: 100%;
+  height: 100%;
+  & .sendbird-thread-ui {
+    max-width: 100%;
+    & .sendbird-thread-ui__header {
+      width: 100%;
+    }
+  }
+}
```
sravan-s, line 19: Shouldnt this better be:

in https://github.com/sendbird/sendbird-uikit-react/blob/feature/UIKIT-3476/Mobile-support-for-thread/src/modules/Thread/components/ThreadUI/index.scss#L3

```
  @include mobile() {
    width: 100%;
  }
```

sravan-s: Ideally, we have to imagine they should work flawlessly even when used independently (ie.outside App(Mobile)).. Given a proper container

src/lib/MediaQueryContext.tsx
HoonBaek: Let's call it `breakpoint`

theajr: @sravan-s while we are here, When I am trying to use `SendBirdApp` or `SendbirdProvider`, Netiher of them seem to respect/accept `mediaQueryBreakPoint` with string or boolean value. I tried passing breakpoint like 400px or true as well. I tried to find example in storybook too, but no luck.

sravan-s: Hey @theajr(I assume that you are using uikit using npm), those props  are not open yet.. Or are you using this branch?

theajr: @sravan-s  I am using v3.4.7, which has `mediaQueryBreakPoint` prop in this version tagged codebase. Do you mean this works only in this branch, not the v3.4.7 yet? Let me know if any help required, will be happy to help.

sravan-s: Ah, so yeah, its not on 3.4.7 ~ 

We are doing QA on it ~ will release as 3.5.0 hopefully around end of May; with docs, tutorials etc~
Thanks for offering help 🙇

theajr: Thank you for clarifying.

sravan-s (approved):

HoonBaek:
I will apply the review in the separate PR

PR #534, opened 2023-05-11
feat: run yarn install only when pkg.json has changes

AhyoungRyu:
### Description Of Changes
Running yarn install on every git pull would be unnecessary. So I added a condition in post-merge hook to check if package.json has been changed. 

https://sendbird.slack.com/archives/C050Z6DA70A/p1683788237121899?thread_ts=1683787988.006319&cid=C050Z6DA70A

### DIFF
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,4 +1,7 @@
 #!/bin/bash
 
-# Install dependencies
-yarn install
+# Run yarn install only when package.json has been changed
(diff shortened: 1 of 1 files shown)


.husky/post-merge
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
 
-# Install dependencies
-yarn install
+# Run yarn install only when there's change in any of below
+# - package.json 
+# - yarn.lock
+# .yarn/*
```
sravan-s, line 6: 👍

sravan-s (approved):

PR #535, opened 2023-05-11
fix: ChannelList blink when when message is send

sravan-s:
Condition to reproduce:
* Implement two channel list(A, B) with different filters
* When a message is send from A, the Channel is shows in B for a second

Reason:
* CHANNEL_REPLACED_TO_TOP is broadcasted to all ChannelLists
* channelListQuery filter was not applied on CHANNEL_REPLACED_TO_TOP reducer 

---
Remove unused reducer SHOW_CHANNEL_SETTINGS, HIDE_CHANNEL_SETTINGS

fixes: https://sendbird.atlassian.net/browse/UIKIT-3824


src/utils/index.ts
```diff
@@ -378,7 +378,7 @@ export const filterChannelListParams = (params: GroupChannelListQuery, channel:
     return false;
   }
   const searchFilter = params?.searchFilter;
-  if (searchFilter?.query && searchFilter.fields?.length > 0) {
+  if (searchFilter?.query && searchFilter.fields && searchFilter.fields.length > 0) {
```
HoonBaek, line 381: nit: how about this?
```suggestion
  if (searchFilter?.query && (searchFilter?.fields?.length ?? 0) > 0) {
```

HoonBaek (approved):
LGTM, especially I love the tests, thank you :D

PR #536, opened 2023-05-15
hot-fix: NodeList conversion on MessageInput.Paste

sravan-s:
NodeList cannot use Array methods, but our conversion was not really explicit.
So, during the ESLint upgrade, NodeList -> Array(Nodes) conversion was deleted

As a fix, we make explicit function to convert NodeList to Array(Nodes)


sravan-s:
Screen Shot 2023-05-15 at 6 33 13 PM

HoonBaek:
Thanks for this fix @sravan-s !

PR #537, opened 2023-05-15
doc: Add info on how to enable Husky

sravan-s:
If husky is not activated using  `yarn husky install`, it wont register :(


AhyoungRyu (approved):
Oh I missed this. Thanks!

PR #538, opened 2023-05-15
fix: prevent white space only text sending

AhyoungRyu:
### Description Of Changes
Resolves https://sendbird.atlassian.net/browse/UIKIT-3869

Previously, even I made the send btn invisible when there's only empty spaces(+ new lines) in the input field, but still the empty white spaces could be sent by hitting an enter button 😅

So I made a fix to block sending like below (see what I'm pressing)
recording

Plus I added more unit tests to prevent any further mistakes.

### DIFF
--- a/src/ui/MessageInput/__tests__/MessageInput.spec.js
+++ b/src/ui/MessageInput/__tests__/MessageInput.spec.js
@@ -59,7 +59,7 @@ describe('ui/MessageInput', () => {
     expect(input.textContent).toBe(mockText);
 
     fireEvent.keyDown(input, { key: 'Enter' });
-    expect(onSendMessage).toHaveBeenCalled();
+    expect(onSendMessage).toHaveBeenCalledWith({ mentionTemplate: mockText, message: mockText });
   });
 
   it('should call sendMessage with valid string; new lines included', async () => {
@@ -74,7 +74,7 @@ describe('ui/MessageInput', () => {
     expect(input.textContent).toBe(mockText);
 
     fireEvent.keyDown(input, { key: 'Enter' });
-    expect(onSendMessage).toHaveBeenCalled();
+    expect(onSendMessage).toHaveBeenCalledWith({ mentionTemplate: mockText, message: mockText });
   });
 
   it('should not call sendMessage with invalid string; only white spaces', async() => {
@@ -89,7 +89,7 @@ describe('ui/MessageInput', () => {
     expect(input.textContent).toBe(mockText);
 
     fireEvent.keyDown(input, { key: 'Enter' });
-    expect(onSendMessage).not.toHaveBeenCalled();
+    expect(onSendMessage).not.toHaveBeenCalledWith({ mentionTemplate: mockText, message: mockText });
   });
   
   it('should render send icon if text is present', async() => {
--- a/src/ui/MessageInput/__tests__/MessageInput.spec.js
+++ b/src/ui/MessageInput/__tests__/MessageInput.spec.js
@@ -1,5 +1,6 @@
 import React, { useRef } from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen,fireEvent, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import MessageInput from "../index";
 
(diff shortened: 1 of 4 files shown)


src/ui/MessageInput/index.jsx
```diff
@@ -137,7 +137,7 @@ const MessageInput = React.forwardRef((props, ref) => {
       //
     }
     setMentionedUserIds([]);
-    setIsInput(textField?.innerText?.trim().length > 0);
+    setIsInput(textField?.textContent?.trim().length > 0);
```
AhyoungRyu, line 140: Unfortunately jsdom doesn't recognize `innerText` So I had to replace it with `textContent` which is  similar to `innerText`  😕 
https://github.com/jsdom/jsdom/issues/1245

Otherwise we won't be able to add unit tests anymore.

sravan-s: give me a little bit of time to test ~

src/ui/MessageInput/__tests__/MessageInput.spec.js
```diff
@@ -45,28 +46,75 @@ describe('ui/MessageInput', () => {
...
-    const component = shallow(
-      <MessageInput
-        onSendMessage={noop}
-        message={{
-          message: 'hello',
-          mentionedMessageTempalte: 'hello'
-        }}
-      />
-    );
+  it('should call sendMessage with valid string', async () => {
```
sravan-s, line 49: 👍

src/ui/MessageInput/__tests__/MessageInput.spec.js
```diff
@@ -45,28 +46,75 @@ describe('ui/MessageInput', () => {
...
+
+    render(<MessageInput onSendMessage={onSendMessage} ref={textRef} />);
+
+    const input = screen.getByRole('textbox');
+    await userEvent.clear(input);
+    await userEvent.type(input, mockText);
+    expect(input.textContent).toBe(mockText);
+
+    fireEvent.keyDown(input, { key: 'Enter' });
+    expect(onSendMessage).toHaveBeenCalled();
```
sravan-s, line 62: nit: Can we do a `toHaveBeenCalledWith` assertion instead?
 https://jestjs.io/docs/expect#tohavebeencalledwitharg1-arg2-

sravan-s: Similar for other tests

src/ui/MessageInput/index.jsx
sravan-s: Need to test `onPaste` condition

AhyoungRyu: How can I test the condition?

src/ui/MessageInput/__tests__/MessageInput.spec.js
AhyoungRyu: Yeah just did https://github.com/sendbird/sendbird-uikit-react/pull/538/commits/c128ae24d0b4531d61fb8e679e277ceb22e32260

src/ui/MessageInput/__tests__/MessageInput.spec.js
```diff
@@ -45,28 +46,75 @@ describe('ui/MessageInput', () => {
     ).toBe(0);
   });
 
-  it.skip('should render send icon if text is present', () => {
```
AhyoungRyu, line 48: Removed the skip tag from here and fixed the test inside here

src/ui/MessageInput/index.jsx
sravan-s: I wanted to do a smoke test manually haha
If you have free time, maybe add a test to detect XSS -> something like https://dev.to/lexlohr/unit-testing-with-jest-typescript-hands-on-session-345o#the-second-task

AhyoungRyu: @sravan-s  So you're trying to test `onPaste` condition -> want me to add a test to detect XSS. I don't get what you really wanted to do 😅 Could you give me more detail what makes you think we need XSS test here?

src/ui/MessageInput/index.jsx
```diff
@@ -420,6 +420,7 @@ const MessageInput = React.forwardRef((props, ref) => {
               e.preventDefault();
             } else {
               if (!e.shiftKey && e.key === MessageInputKeys.Enter
+                && textField?.textContent?.trim().length > 0
```
AhyoungRyu, line 423: This newly added line is the only one directly related to "prevent white spaces only sending by hitting the enter key"

src/ui/MessageInput/index.jsx
AhyoungRyu: Had a short huddle with Sravan 🎧 
 - XSS test is not related to this change but kind of suggestion which is  good to have: probably will do separately
 - There could be some edge cases hardly caught by unit tests so manual tests is also needed

sravan-s (approved):

sravan-s (approved):
Tested manually ~ looks good

AhyoungRyu:
@sravan-s Oh I manually tested from my side too but there's an issue when I try to mention someone's id. It throws an runtime error and it seemed related to this change. To fix this issue, I could modify other part of code but I thought it would be still risky. So I put the `innerText` back and only the part of that I newly added is using `textContent` which doesn't impact on the existing features. https://github.com/sendbird/sendbird-uikit-react/pull/538/commits/63c9fdd4208567da73ac10e0eb64962d9d652d3c 

Let a comment too.

src/ui/MessageInput/__tests__/MessageInput.spec.js
```diff
@@ -92,7 +92,7 @@ describe('ui/MessageInput', () => {
     expect(onSendMessage).not.toHaveBeenCalledWith({ mentionTemplate: mockText, message: mockText });
   });
   
-  it('should render send icon if text is present', async() => {
```
AhyoungRyu: Unfortunately had to put the skip tag again 🥲

src/ui/MessageInput/index.jsx
AhyoungRyu: Attaching this link too https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext for someone who might wonder

PR #547, opened 2023-05-19
Fix: Scroll towards top

sravan-s:
Scroll top in MessageList

### DIFF
--- a/src/modules/Channel/components/MessageList/index.tsx
+++ b/src/modules/Channel/components/MessageList/index.tsx
@@ -70,7 +70,7 @@ const MessageList: React.FC<MessageListProps> = ({
       scrollHeight,
     } = element;
 
-    if (isAboutSame(clientHeight + scrollTop, scrollHeight, SCROLL_BUFFER)) {
+    if (isAboutSame(scrollTop, 0, SCROLL_BUFFER)) {
       onScrollDownCallback(([messages]) => {
         if (messages) {
           try {


HoonBaek (approved):

src/hooks/useHandleOnScrollCallback/index.ts
```diff
@@ -38,19 +43,24 @@ export function useHandleOnScrollCallback({
...
+    if (hasMore && scrollTop < SCROLL_BUFFER) {
       onScroll(() => {
         // sets the scroll position to the bottom of the new messages
         element.scrollTop = element.scrollHeight - scrollBottom;
       });
     }
+    if (hasNext) {
+      onScroll(() => {
+        // sets the scroll position to the top of the new messages
+        element.scrollTop = scrollTop - (scrollHeight - element.scrollHeight);
```
HoonBaek, line 55: Oh, this is what I wanted to do 👍

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -116,6 +130,7 @@ const MessageList: React.FC<MessageListProps> = ({
 
   const handleOnScroll = useHandleOnScrollCallback({
     hasMore: hasMorePrev,
+    hasNext: hasMoreNext,
```
HoonBaek, line 133: Could you leave a comment here?
like
```
// TODO: rename these two props, hasMore & hasNext
```

HoonBaek: Nah, let's do not that for now, I can make fix after release

PR #548, opened 2023-05-22
feat: add snakeToCamel util function

AhyoungRyu:
### Description Of Changes
Applies https://sendbird.atlassian.net/browse/UIKIT-3974

The format of the UIKitConfiguration payload will consist of snake_case like [this](https://sendbird.atlassian.net/wiki/spaces/SDK/pages/2060618878/UIKit+Feature+configuration+RFC#JSON-Payload). So I made a utility function to convert it into camelCase, which is more natural in our application."

### DIFF
--- a/src/utils/__tests__/snakeToCamel.spec.ts
+++ b/src/utils/__tests__/snakeToCamel.spec.ts
@@ -1,85 +1,88 @@
 import { snakeToCamel } from '../snakeToCamel';
 
+// mock uikit configurations
+const mockSnakeCase = {
+  last_updated_at: 1234567890123,
+  uikit_configurations: {
+    common: {
+      enable_using_default_user_profile: false,
+    },
+    group_channel: {
+      channel: {
+        enable_ogtag: true,
+        enable_typing_indicator: true,
+        enable_reactions: true,
+        enable_mention: false,
+        enable_voice_message: false,
+        reply_type: 'quote_reply',
+        thread_reply_select_type: 'thread',
+        input: {
+          camera: {
+            enable_photo: true,
+            enable_video: true,
+          },
+          gallery: {
+            enable_photo: true,
+            enable_video: true,
+          },
+          enable_document: true,
+        },
+      },
+      channel_list: {
+        enable_typing_indicator: false,
+        enable_message_receipt_status: false,
+      },
+      setting: {
+        enable_message_search: false,
+      },
+    },
+  },
+};
+
+const mockCamelCase = {
+  lastUpdatedAt: 1234567890123,
+  uikitConfigurations: {
+    common: {
+      enableUsingDefaultUserProfile: false,
+    },
+    groupChannel: {
+      channel: {
+        enableMention: false,
+        enableOgtag: true,
+        enableReactions: true,
+        enableTypingIndicator: true,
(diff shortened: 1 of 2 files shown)


sravan-s (approved):

src/utils/__tests__/snakeToCamel.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    expect(snakeToCamel(input)).toEqual(expectedOutput);
+  });
+
+  it('should handle empty objects and arrays', () => {
+    expect(snakeToCamel({})).toEqual({});
+    expect(snakeToCamel([])).toEqual([]);
+  });
+
+  it('should preserve non-object and non-array values', () => {
+    expect(snakeToCamel('test')).toBe('test');
```
sravan-s, line 91: suggestion: good to add test for `camel_case`,

src/utils/snakeToCamel.ts
```diff
@@ -0,0 +1,29 @@
+type CamelCase<T> = T extends object
```
sravan-s, line 1: 👍

src/utils/__tests__/snakeToCamel.spec.ts
AhyoungRyu: Yeah I added it [here](https://github.com/sendbird/sendbird-uikit-react/pull/548/commits/53a68d8b91289f258aa0b88c884c75df5cd0d3a7#diff-f73dfe797e2645768392a250b19223f6cca5439911d3235815b4eb37e7f97cadR84).

src/utils/snakeToCamel.ts
```diff
@@ -0,0 +1,29 @@
...
+    return obj as CamelCase<T>;
+  }
+
+  if (Array.isArray(obj)) {
+    return obj.map((item) => snakeToCamel(item)) as CamelCase<T>;
+  }
+
+  const result = {} as CamelCase<T>;
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
HoonBaek, line 22: Do we need to use the prototype function here? because It already checked the types above
(just curious)

HoonBaek (approved):
LGTM

src/utils/snakeToCamel.ts
AhyoungRyu: @HoonBaek Where's the above?

HoonBaek: on the line numbers 12 and 16
Isn't already checked?

AhyoungRyu: Ahh this line is for checking if the object has the key not checking if its type is Object or Array. 

Also, this line might look unnecessary cause the `key` is from just its above like in for-in loop. 
You can find the detail of behind reason is described in [this Stackoverflow question](https://stackoverflow.com/questions/1963102/what-does-the-jslint-error-body-of-a-for-in-should-be-wrapped-in-an-if-statemen). 

But anyways, I also didn't like this pattern either. So I dropped the for-in + if pattern and replaced it \w `Object.keys.forEach(key => )`. 992def4bce33a54292532cc55703275e2c85a02a 

Thanks for mentioning it Hoon :)

PR #554, opened 2023-05-24
fix: display a channel only when there's a msg

AhyoungRyu:
### Description Of Changes
Resolves https://sendbird.atlassian.net/browse/UIKIT-3969 

A newly created channel should be displayed on the invitee's channel list only when there's a message not right after the channel is created.


HoonBaek (approved):
Thank you!

PR #559, opened 2023-05-25
fix: Inconsistent rendering on scrollToBottom button

sravan-s:
While refactor, we removed scrollToButton, it got coupled with `handleOnScroll`
which means, it would be called only when data fetching is involved
Seperating these logics would fix this

fixes: https://sendbird.atlassian.net/browse/UIKIT-3972

### DIFF
--- a/src/modules/Channel/components/MessageList/index.tsx
+++ b/src/modules/Channel/components/MessageList/index.tsx
@@ -19,7 +19,7 @@ import { MessageProvider } from '../../../Message/context/MessageProvider';
 import { useHandleOnScrollCallback } from '../../../../hooks/useHandleOnScrollCallback';
 import { useSetScrollToBottom } from './hooks/useSetScrollToBottom';
 
-const SCROLL_BOTTOM_PADDING = 200;
+const SCROLL_BOTTOM_PADDING = 50;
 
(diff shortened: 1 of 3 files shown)


sravan-s:
Kapture 2023-05-25 at 23 17 15

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -17,6 +17,9 @@ import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
 import { UserMessage } from '@sendbird/chat/message';
 import { MessageProvider } from '../../../Message/context/MessageProvider';
 import { useHandleOnScrollCallback } from '../../../../hooks/useHandleOnScrollCallback';
+import { useSetScrollToBottom } from './hooks/useSetScrollToBottom';
+
+const SCROLL_BOTTOM_PADDING = 200;
```
HoonBaek, line 22: isn't it too big?

HoonBaek (approved):
Wow, the test codes look awesome

src/modules/Channel/components/MessageList/index.tsx
sravan-s: set as 50

PR #551, opened 2023-05-23
feat: save uikitConfiguration in global state

AhyoungRyu:
### Description Of Changes
Applies https://sendbird.atlassian.net/browse/UIKIT-3934

 - Updated minimum required `@sendbird/chat` version to 4.8.3 which includes [the UIKitConfiguration feature](https://github.com/sendbird/chat-js/pull/1131) in chat-js. 
 - Saved the response of `sdk.getUIKitConfiguration` in Global state which is the same level with sdk & user. To achieve this, added dux/uikitConfiguration as well.

PR #561, opened 2023-05-26
feat: Add basic implementation of UIKitConfigProvider

AhyoungRyu:
### Description Of Changes
 - [x] Addressed partial implementation in https://sendbird.atlassian.net/browse/UIKIT-4014?focusedCommentId=218951 commented by @bang9 🎉 
 - [x] Removed dux/uikitConfiguration which had been added by https://github.com/sendbird/sendbird-uikit-react/pull/551 to replace it \w UIkitConfigProvider


src/lib/UIKitConfigProvider/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+  },
+};
+
+interface UIKitConfigContextInterface {
+  initDashboardConfigs: (sdk: SendbirdChat) => void;
+  configs: UIKitConfigInfo;
+}
+const UIKitConfigContext = createContext<UIKitConfigContextInterface>({
+  /* eslint-disable @typescript-eslint/no-empty-function */
+  initDashboardConfigs: () => {},
```
bang9, line 67: we can use `noop` instead of /* eslint-disable */
https://github.com/sendbird/sendbird-uikit-react/blob/main/src/utils/utils.js#L1

src/lib/UIKitConfigProvider/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      // if(sdk.appInfo.uikitConfigInfo.lastUpdatedAt === storedConfigs.lastUpdatedAt) {
+      //   setRemoteConfigs(storedConfigs);
+      // } else {
+      //   const newConfigs = await sdk.getUIKitConfig();
+      //   const updatedConfigs = await manager.update(newConfigs);
+      //   setRemoteConfigs(updatedConfigs);
+      // }
+      onSuccess?.();
+    } catch (error) {
+      onError?.(error);
```
bang9, line 108: It seems a bit awkward to receive the completion handler of a function executed in a different context as props and handle it. Wouldn't it be better to directly handle it when calling the function? (then/catch, try/catch, or callback)

src/lib/UIKitConfigProvider/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const initDashboardConfigs = useCallback(async (sdk: SendbirdChat) => {
+    try {
+    // TODO: Implement
+    // const manager = new UIKitConfigStorageManager(storage)
+    // const storedConfigs = await manager.init(sdk.appId);
+
+      if (storage == null) {
+        const payload = await sdk.getUIKitConfiguration().json;
+        setRemoteConfigs(snakeToCamel(payload.last_updated_at));
```
bang9, line 94: maybe `snakeToCamel(payload.uikit_configurations)` ?

src/lib/UIKitConfigProvider/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+};
+
+interface UIKitConfigContextInterface {
+  initDashboardConfigs: (sdk: SendbirdChat) => void;
+  configs: UIKitConfigInfo;
+}
+const UIKitConfigContext = createContext<UIKitConfigContextInterface>({
+  /* eslint-disable @typescript-eslint/no-empty-function */
+  initDashboardConfigs: () => {},
+  configs: {} as UIKitConfigInfo,
```
bang9, line 68:
```suggestion
  configs: initialConfig
```

src/lib/UIKitConfigProvider/types.ts
```diff
@@ -20,7 +22,7 @@ interface GroupChannelConfig {
         enableVideo: boolean;
       };
     };
-    replyType: 'none' | 'quote_reply' | 'thread';
+    replyType: CaseInsensitive<'none' | 'quote_reply' | 'thread'>;
```
bang9, line 25: Question, why is this needed?

AhyoungRyu: Ahh we already have replyType which has all capital letters inside. https://github.com/sendbird/sendbird-uikit-react/blob/main/src/types.d.ts#L9. So I wanted to make it accept lowercase and capital letters both. 

I should've left comment about it. Just updated https://github.com/sendbird/sendbird-uikit-react/pull/561/commits/27596f85bbae94a78a3facff2cb9672d9e969400

src/lib/UIKitConfigProvider/index.tsx
AhyoungRyu: Yeah my bad 😅
Fixed in https://github.com/sendbird/sendbird-uikit-react/pull/561/commits/5245991197cc8ef39fd7dfd14086c2d9474679da#diff-7be02629955aad804d99af819265cafc611c0fde160e3eb739e0988c80ac3e4dR91

src/lib/UIKitConfigProvider/index.tsx
AhyoungRyu: Yeah agreed. Here's the change :) 
https://github.com/sendbird/sendbird-uikit-react/pull/561/commits/5245991197cc8ef39fd7dfd14086c2d9474679da

src/lib/UIKitConfigProvider/index.tsx
AhyoungRyu: Its type was wrong so fixed the type and also removed the disabling comment. https://github.com/sendbird/sendbird-uikit-react/pull/561/commits/5245991197cc8ef39fd7dfd14086c2d9474679da#diff-7be02629955aad804d99af819265cafc611c0fde160e3eb739e0988c80ac3e4dL66

bang9: 🙆‍♂️

src/lib/Sendbird.tsx
```diff
@@ -242,18 +253,32 @@ const Sendbird = ({
...
           voiceRecord,
           userMention: {
             maxMentionCount: userMention?.maxMentionCount || 10,
             maxSuggestionCount: userMention?.maxSuggestionCount || 15,
           },
-          isTypingIndicatorEnabledOnChannelList,
-          isMessageReceiptStatusEnabledOnChannelList,
-          replyType,
           markAsReadScheduler,
+          ...uikitConfigurations,
```
sravan-s, line 262: Do we need this here with `SendbirdSdkContext`?

src/lib/UIKitConfigProvider/utils/types.ts
```diff
@@ -0,0 +1,13 @@
+// @link https://gist.github.com/navix/6c25c15e0a2d3cd0e5bce999e0086fc9
+/* eslint-disable @typescript-eslint/ban-types */
+export type DeepPartial<T> = T extends Function
```
sravan-s, line 3: Can we move this to `typeHelpers`

AhyoungRyu: Ah I should've mention about it too. I'll move the whole logic under `lib/UIKitConfigProvider` to https://github.com/sendbird/sendbird-uikit-core-ts/tree/main/packages/uikit-tools so RN and React can share the logic. What do you think?

src/lib/UIKitConfigProvider/types.ts
bang9: Ah~ I understand your point.

But one opinion is that it is preferable to maintain the original form of the API response as much as possible. Changing the snake_case may not be a significant issue, but modifying the type of values can lead developers further away from their initial expectations and impose mandatory tracking of history.

Moreover, when using the UI for rendering, it seems more beneficial in terms of maintainability and readability to map the values only once when passing them as props to UIKitConfigProvider, rather than performing checks for the two types every time.

```ts
if (config.replyType === 'thread' || config.replyType === 'THREAD') { ... } // x Component count

// vs

config={{ replyType: mapToReplyType('THREAD') }} 
if (config.replyType === 'thread') { ... } // x Component count
```

In practice, since only 'THREAD' is used in react-uikit, it is possible to perform the if statement check only for 'THREAD'. However, this can impose the need for history tracking on the person reading the code once again.

src/lib/Sendbird.tsx
AhyoungRyu: I left the configs which is being passed to SendbirdSdkContext.Provider for smooth migration. 
Is this what you're asking?

bang9: IMHO, To maintain the migration by keeping the config of sdkContext in view, it seems safer to elevate UIKitConfigProvider to the top level and pass the values set in UIKitConfigProvider to SendbirdSdkContext through a consumer.

src/lib/UIKitConfigProvider/utils/types.ts
sravan-s: Cool

src/lib/Sendbird.tsx
sravan-s: What is smoother migration in this situation, btw .. sorry, I am a bit off "context" :D

AhyoungRyu: Oh sorry I should've provided more context 😅

So currently we're passing the app level of config to SendbirdSDK provider. But the thing I'm trying to do in this PR is making UIKitConfigProvider handles everything(most of things) related feature configuration instead of SendbirdSDK. 

But I didn't want to make any breaking changes by removing the config values from SendbirdSDK at once. Smoother migration is about this. 

Not sure I gave you a better context ㅋㅋ

src/lib/UIKitConfigProvider/types.ts
AhyoungRyu: @bang9 Yeah I see your point. 
I removed the custom CaseInsensitive type from UIKitConfigurationProvider. Instead, I convert the replyType config value from where it's used. 
https://github.com/sendbird/sendbird-uikit-react/pull/561/commits/98890740b5f158f2643b8c8a21c076c6c62727fd 

Instead of creating a separate util like mapToRelyType, just used toLowerCase(). But feel free to tell me if you see concern :)

bang9:
혹시 요 PR 의 작업범위가 어떻게 되는걸까요? UIKitConfigProvider 에서 받아온 Dashboard configs 는 아직 UI 에 반영이 안되어야 하는게 맞을까용?

AhyoungRyu:
>혹시 요 PR 의 작업범위가 어떻게 되는걸까요? UIKitConfigProvider 에서 받아온 Dashboard configs 는 아직 UI 에 반영이 안되어야 하는게 맞을까용?

네네 맞습니다 UIKitConfigProvider 를 어떻게 구성할지 대략적인 구현 추가해봤고, 또 uikit-react 에선 어느 위치에서 사용할지 적용해보는 정도로만 구현해봤습니다

bang9 (approved):

sravan-s (approved):
LGTM, in commit, will be good if you can include https://github.com/sendbird/sendbird-uikit-react/pull/561#issuecomment-1564425849 on squashed commit ~

AhyoungRyu:
It's been merged. Will continue to implement the missing part one by one in UIKitProvider which has "TODO: implement" comment :)  \cc @bang9

src/lib/Sendbird.tsx
AhyoungRyu: @bang9 Sorry I missed [your comment](https://github.com/sendbird/sendbird-uikit-react/pull/561#discussion_r1206285328) at the above about the consumer. Just created a PR for it here https://github.com/sendbird/sendbird-uikit-react/pull/572 :)

PR #562, opened 2023-05-26
fix: Display menu for open channel OG message

HoonBaek:
### Description Of Changes

Move applying onLongPress events to the top level element

[UIKIT-4001](https://sendbird.atlassian.net/browse/UIKIT-4001)

[UIKIT-4001]: https://sendbird.atlassian.net/browse/UIKIT-4001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/OpenchannelOGMessage/index.tsx
+++ b/src/ui/OpenchannelOGMessage/index.tsx
@@ -61,27 +61,32 @@ export default function OpenchannelOGMessage({
   userId,
 }: OpenChannelOGMessageProps): JSX.Element {
   if (!message || message.messageType !== 'user') {
-    return null;
+    return <></>;
   }
 
+  const status = message?.sendingStatus;
+  const ogMetaData = message.ogMetaData ?? null;
+  const defaultImage = ogMetaData?.defaultImage;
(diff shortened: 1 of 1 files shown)


src/ui/OpenchannelOGMessage/index.tsx
```diff
@@ -61,27 +61,32 @@ export default function OpenchannelOGMessage({
   userId,
 }: OpenChannelOGMessageProps): JSX.Element {
   if (!message || message.messageType !== 'user') {
-    return null;
+    return <></>;
```
sravan-s, line 64: This return here is not safe. it would possibly break rule of hooks(very improbable)
, better move to it to line 113 (right before the final return)

sravan-s (approved):

src/ui/OpenchannelOGMessage/index.tsx
sravan-s: !important ^^

sravan-s (approved):

PR #564, opened 2023-05-26
fix: Un-expose reactions menu in special channels

HoonBaek:
### Description Of Changes

* Do not show emoji reactions in the Supergroup channel & Broadcast group channel
  * Create a util func: getIsReactionEnabled
  * Use the function to everywhere calculate the condition of isReactionEnabled

[UIKIT-4008](https://sendbird.atlassian.net/browse/UIKIT-4008)

[UIKIT-4008]: https://sendbird.atlassian.net/browse/UIKIT-4008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Sendbird.tsx
+++ b/src/lib/Sendbird.tsx
@@ -29,6 +29,7 @@ import { VOICE_RECORDER_DEFAULT_MAX, VOICE_RECORDER_DEFAULT_MIN } from '../utils
 import { useMarkAsReadScheduler } from './hooks/useMarkAsReadScheduler';
 import { useMarkAsDeliveredScheduler } from './hooks/useMarkAsDeliveredScheduler';
 import { ConfigureSessionTypes } from './hooks/useConnect/types';
+import { getIsReactionEnabled } from '../utils/getIsReactionEnabled';
 
 export type UserListQueryType = {
   hasNext?: boolean;
@@ -213,7 +214,10 @@ const Sendbird = ({
     // common.enable_using_default_user_profile
     disableUserProfile,
     // group_channel.enable_reactions
-    isReactionEnabled,
+    isReactionEnabled: getIsReactionEnabled({
+      appLevel: sdkStore?.sdk?.appInfo?.useReaction,
+      globalLevel: isReactionEnabled,
+    }),
     // group_channel.enable_mention
     isMentionEnabled: isMentionEnabled || false,
     // group_channel.enable_voice_message
--- a/src/lib/Sendbird.tsx
+++ b/src/lib/Sendbird.tsx
@@ -29,7 +29,6 @@ import { VOICE_RECORDER_DEFAULT_MAX, VOICE_RECORDER_DEFAULT_MIN } from '../utils
 import { useMarkAsReadScheduler } from './hooks/useMarkAsReadScheduler';
 import { useMarkAsDeliveredScheduler } from './hooks/useMarkAsDeliveredScheduler';
 import { ConfigureSessionTypes } from './hooks/useConnect/types';
-import { getIsReactionEnabled } from '../utils/getIsReactionEnabled';
 
 export type UserListQueryType = {
   hasNext?: boolean;
@@ -214,10 +213,7 @@ const Sendbird = ({
     // common.enable_using_default_user_profile
     disableUserProfile,
     // group_channel.enable_reactions
-    isReactionEnabled: getIsReactionEnabled({
-      appLevel: sdkStore?.sdk?.appInfo?.useReaction,
-      globalLevel: isReactionEnabled,
-    }),
+    isReactionEnabled: isReactionEnabled,
     // group_channel.enable_mention
     isMentionEnabled: isMentionEnabled || false,
     // group_channel.enable_voice_message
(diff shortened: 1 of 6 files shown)


src/lib/Sendbird.tsx
```diff
@@ -242,7 +243,10 @@ const Sendbird = ({
             compressionRate: 0.7,
             ...imageCompression,
           },
-          isReactionEnabled,
+          isReactionEnabled: getIsReactionEnabled({
+            appLevel: sdkStore?.sdk?.appInfo?.useReaction,
+            globalLevel: isReactionEnabled,
+          }),
```
HoonBaek, line 249: @AhyoungRyu This is related to the feature configuration, I think

AhyoungRyu: What's the difference between appLevel <-> globalLevel here?

src/utils/getIsReactionEnabled.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export function getIsReactionEnabled({
+  appLevel = true,
+  isBroadcast = false,
+  isSuper = false,
+  globalLevel = true,
+  moduleLevel = true,
+}: IsReactionEnabledProps): boolean {
+  return appLevel && !isBroadcast && !isSuper && globalLevel && moduleLevel;
```
bang9, line 21: global level -> false
module level -> true

인 경우에는 true 로 리턴이 되어야 할 것 같습니당

src/lib/Sendbird.tsx
HoonBaek: appLevel means the value which is set remotely (ex. from dash board)
and globalLevel means it's set through SendbirdProvider

src/utils/getIsReactionEnabled.ts
```diff
@@ -0,0 +1,31 @@
...
+  globalLevel = true,
+  moduleLevel = null,
+}: IsReactionEnabledProps): boolean {
+  if (moduleLevel) {
+    return !isBroadcast && !isSuper;
+  }
+  return (
+    appLevel
+    && globalLevel
+    && (moduleLevel ?? true)
```
AhyoungRyu, line 27: Q. what's the reason behind we set the fallback value of moduleLevel as true here?

src/utils/getIsReactionEnabled.ts
```diff
@@ -0,0 +1,31 @@
...
+ * This function helps consider the every condition
+ * related to enabling emoji reaction feature.
+ */
+
+export interface IsReactionEnabledProps {
+  appLevel?: boolean;
+  isBroadcast?: boolean;
+  isSuper?: boolean;
+  globalLevel?: boolean;
+  moduleLevel?: boolean | null;
```
AhyoungRyu, line 11: Is there any specific reason we allow null for moduleLevel only?

src/lib/Sendbird.tsx
AhyoungRyu: Do we have more configs under appInfo like `useReaction`? If so, should we handle other configs in appInfo too in this way?

HoonBaek: No, as Airen said we don't need to consider the appInfo anymore.
I will remove the line

src/utils/getIsReactionEnabled.ts
HoonBaek: When moduleLevel is undefined so it's null, it shouldn't be falsy

src/utils/getIsReactionEnabled.ts
HoonBaek: Because I set the default value with `null`.
Do you think it's better to keep it undefined? then I will change

AhyoungRyu: I thought it'd be better if we can set its default value \w false or true as other fields do.

AhyoungRyu (approved):
👍👍👍

src/utils/getIsReactionEnabled.ts
```diff
@@ -0,0 +1,28 @@
...
+  globalLevel?: boolean;
+  moduleLevel?: boolean;
+}
+
+export function getIsReactionEnabled({
+  isBroadcast = false,
+  isSuper = false,
+  globalLevel = true,
+  moduleLevel,
+}: IsReactionEnabledProps): boolean {
```
bang9, line 18: 로직을 조금 간소화 할 수 있을것 같습니다.

p0. isBroadcast / isSuper -> level 에 상관없이 해당하면 false 로 떨어져야 합니다.
p1. moduleLevel -> globalLevel 에 상관없이, 값이 주어지면 해당 값으로 떨어져야 합니다.
p2. globalLevel -> p0/p1 에 해당하지 않으면, 해당 값으로 떨어져야 합니다.

HoonBaek: 오오.. 감사합니다!! 코드가 훨씬 간결해 질 것 같습니다

PR #566, opened 2023-05-28
fix: Improve max mention count logic 

sravan-s:
Array of Users is de-duplicated, so max-mention count wouldnt work,
So, use DOM mutations to reliabily count correct no.of mentioned users

Bonus:
 Improve MessageProvider type defn

Fixes: https://sendbird.atlassian.net/browse/UIKIT-3982

### DIFF
--- a/src/modules/Message/hooks/useDirtyGetMentions.ts
+++ b/src/modules/Message/hooks/useDirtyGetMentions.ts
@@ -37,11 +37,9 @@ export function useDirtyGetMentions({
 
     // Callback function to execute when mutations are observed
     const callback = (mutationList: MutationRecord[]): void => {
-      for (const mutation of mutationList) {
-        if (mutation.type === 'childList') {
-          // setMentionNodes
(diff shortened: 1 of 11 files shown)


src/modules/Message/hooks/useDirtyGetMentions.ts
```diff
@@ -0,0 +1,67 @@
...
+  useEffect(() => {
+    // Options for the observer (which mutations to observe)
+    const config = { childList: true, subtree: true };
+
+    // Callback function to execute when mutations are observed
+    const callback = (mutationList: MutationRecord[]): void => {
+      for (const mutation of mutationList) {
+        if (mutation.type === 'childList') {
+          // setMentionNodes
+          setMentionNodes(getMentionNodes(targetNode));
```
sravan-s, line 43: Note to self -> debounce, also move out of loop

HoonBaek (approved):

PR #573, opened 2023-05-30
hotfix: disableMarkAsRead should be passed

sravan-s:
disableMarkAsRead should be passed down as a param in useHandleChannelEvents


HoonBaek (approved):
LGTM

PR #598, opened 2023-06-02
feat: Modify replyType & isVoiceMessageEnabled default value

AhyoungRyu:
### Description Of Changes
Addresses https://sendbird.atlassian.net/browse/UIKIT-4082

Changed default prop values
1 `isVoiceMessageEnabled` : true → false
2. `replyType`: ‘none’ → 'quote_reply’


src/lib/Sendbird.tsx
```diff
@@ -91,13 +91,13 @@ export interface SendbirdProviderProps extends CommonUIKitConfigProps {
 
 function Sendbird(props: SendbirdProviderProps) {
   const {
-    replyType = 'NONE',
-    isMentionEnabled = false,
-    isReactionEnabled = true,
-    disableUserProfile = false,
-    isVoiceMessageEnabled = true,
-    isTypingIndicatorEnabledOnChannelList = false,
-    isMessageReceiptStatusEnabledOnChannelList = false,
```
AhyoungRyu, line 100: Don't do this anymore in here since this(it's P3) will break the config priority rule.

```
- 
    - **[P3]** UIKit default value
    - **[P2]** Values set from Dashboard
    - **[P1]** Values set in code by the customer
```

By removing this P3 default value setting logic, we make P2 config value beat P3 via UIKitConfigProvider.

src/modules/App/index.jsx
```diff
@@ -158,24 +158,24 @@ App.defaultProps = {
...
   isMessageGroupingEnabled: true,
   stringSet: null,
   colorSet: null,
   imageCompression: {},
   disableAutoSelect: false,
+  isMentionEnabled: false,
+  isReactionEnabled: true,
+  replyType: 'QUOTE_REPLY',
+  disableUserProfile: false,
+  isVoiceMessageEnabled: false,
```
AhyoungRyu, line 178: Just put every common configs \w dashboard ones together to see at a glance. 
Only replyType & isVoiceMessageEnabled value has been changed as I wrote in the description :)

src/modules/App/stories/index.stories.js
```diff
@@ -275,6 +275,7 @@ export const user1 = () => fitPageSize(
     replyType="THREAD"
     isReactionEnabled
     isMentionEnabled
+    isVoiceMessageEnabled
```
AhyoungRyu, line 278: Added this prop just in case we're confused in storybook on testing.

sravan-s (approved):

PR #604, opened 2023-06-05
hotfix: make replyType case conversion safe

AhyoungRyu:
### Description Of Changes
(Reported directly by Sravan 😄)

Screen Shot 2023-06-05 at 4 41 06 PM

`replyType` prop barely has null value but there's a chance to be null cause we allow it. So making it safe before it goes into `toLowerCase()`.

### DIFF
--- a/src/lib/Sendbird.tsx
+++ b/src/lib/Sendbird.tsx
@@ -21,6 +21,9 @@ import pubSubFactory from './pubSub/index';
 import useAppendDomNode from '../hooks/useAppendDomNode';
 
 import { UIKitConfigProvider, useUIKitConfig } from './UIKitConfigProvider';
+import initialConfig from './UIKitConfigProvider/const/initialConfig';
+import { GroupChannelConfig } from './UIKitConfigProvider/types';
+
 import { VoiceMessageProvider } from './VoiceMessageProvider';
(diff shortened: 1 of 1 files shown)


sravan-s (approved):
THnx 👍

src/lib/Sendbird.tsx
```diff
@@ -116,7 +119,9 @@ function Sendbird(props: SendbirdProviderProps) {
              * we convert it from here.
              * i.e. 'THREAD' -> 'thread'
              */
-            replyType: replyType.toLowerCase(),
+            replyType: replyType == null
+              ? initialConfig.groupChannel.channel.replyType
+              : replyType.toLowerCase() as GroupChannelConfig['channel']['replyType'],
```
sravan-s, line 124: Probably better to be moved out to a function, as discussed in older PR , but is not a dealbreaker

AhyoungRyu: Oh probably I missed the comment that you commented earlier. Do u remember what you suggested?

sravan-s: No, I thiknk its @bang9  suggested it

AhyoungRyu: I see. Let me merge this one first (your dev env is broken, right?) and look the comment up.

PR #608, opened 2023-06-06
fix: Race condition in playing audio file

sravan-s:
Race condition occurs when two audio files are requested from
backend clear previous player and play only the last arrived file

Cleanup - Do not need to fetch audio file if its already in cache

Fixes: https://sendbird.atlassian.net/browse/UIKIT-3952

### DIFF
--- a/src/hooks/VoicePlayer/index.tsx
+++ b/src/hooks/VoicePlayer/index.tsx
@@ -94,19 +94,25 @@ export const VoicePlayerProvider = ({
     }
 
     logger.info('VoicePlayer: Start getting audio file.');
-    new Promise((resolve) => {
-      if (audioFile) {
-        resolve(audioFile);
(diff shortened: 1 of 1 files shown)


src/hooks/VoicePlayer/index.tsx
```diff
@@ -94,19 +94,25 @@ export const VoicePlayerProvider = ({
...
+      voicePlayerDispatcher({
+        type: INITIALIZE_AUDIO_UNIT,
+        payload: { groupKey },
+      });
+      // audio file passed as a parameter
       if (audioFile) {
-        resolve(audioFile);
         logger.info('VoicePlayer: Use the audioFile instance.');
+        resolve(audioFile);
+        return;
```
sravan-s, line 106: if there is audioFile, no need to fetch from backend

PR #610, opened 2023-06-06
fix: OpenChannel Context menu click leak

sravan-s:
In Mobile,
Touch from contextMenus were getting leaked to MessageBody
To avoid this, move context menu outside of MessageBody's JSX

```
<div longPress>
	<Message />
	<Menu/>
</div>
```

to 
```
<>
	<div longPress>
		<Message />
	</div>
	{
		<Menu/>
	}
</>
```

Bonus * OGMessage should fallback to text-messages

Fixes: https://sendbird.atlassian.net/browse/UIKIT-4029

### DIFF
--- a/src/modules/OpenChannel/components/OpenChannelMessage/utils.ts
+++ b/src/modules/OpenChannel/components/OpenChannelMessage/utils.ts
@@ -17,9 +17,17 @@ export const SendingMessageStatus = {
   PENDING: 'pending',
 };
 
-export const getMessageType = (message: UserMessage | FileMessage | AdminMessage): string => {
+type MessageTypeOptions = {
+  isOgMessageEnabledInOpenChannel?: boolean;
+};
+
+export const getMessageType = (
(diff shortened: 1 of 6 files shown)


src/modules/OpenChannel/components/OpenChannelMessage/utils.ts
```diff
@@ -17,9 +17,17 @@ export const SendingMessageStatus = {
...
-export const getMessageType = (message: UserMessage | FileMessage | AdminMessage): string => {
+type MessageTypeOptions = {
+  isOgMessageEnabledInOpenChannel?: boolean;
+};
+
+export const getMessageType = (
+  message: UserMessage | FileMessage | AdminMessage,
+  options?: MessageTypeOptions,
+): string => {
+  const isOgMessageEnabledInOpenChannel = options?.isOgMessageEnabledInOpenChannel;
```
sravan-s, line 28: cc @AhyoungRyu , its better to fallback to text message when og message is not present

PR #612, opened 2023-06-07
hotfix: modify uikit config payload format

AhyoungRyu:
### Description Of Changes
Currently storybook is throwing a runtime error due to the wrong json format access which is caused by my prev PR.

### DIFF
--- a/src/lib/UIKitConfigProvider/index.tsx
+++ b/src/lib/UIKitConfigProvider/index.tsx
@@ -36,7 +36,7 @@ const UIKitConfigProvider = ({ storage, children, localConfigs }: UIKitConfigPro
     try {
       if (storage == null) {
         const payload = await sdk.getUIKitConfiguration();
-        setRemoteConfigs(snakeToCamel(payload.json));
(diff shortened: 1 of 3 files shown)


src/lib/UIKitConfigProvider/index.tsx
```diff
@@ -36,7 +36,7 @@ const UIKitConfigProvider = ({ storage, children, localConfigs }: UIKitConfigPro
     try {
       if (storage == null) {
         const payload = await sdk.getUIKitConfiguration();
-        setRemoteConfigs(snakeToCamel(payload.json));
+        setRemoteConfigs(snakeToCamel(payload.json).configuration.uikitConfigurations);
```
AhyoungRyu, line 39: I shouldn't have removed `uikitConfigurations` this one but I guess it's removed by mistake 😅

AhyoungRyu: Also, the json data format has been changed in the latest version of sendbird/chat; `configuration` is added in the root level of payload.

sravan-s (approved):
THnx

sravan-s (approved):
THnx

PR #614, opened 2023-06-07
feat: replace local uikitConfigProvider \w uikit-tools one

AhyoungRyu:
### Description Of Changes
Drop the whole logic under `lib/UIKitConfigProvider/*` and replace the logic with the same one in `@sendbird/uikit-tools` made by https://github.com/sendbird/sendbird-uikit-react-native/pull/94.


src/lib/Sendbird.tsx
```diff
@@ -121,7 +119,7 @@ function Sendbird(props: SendbirdProviderProps) {
              */
             replyType: replyType == null
               ? initialConfig.groupChannel.channel.replyType
-              : replyType.toLowerCase() as GroupChannelConfig['channel']['replyType'],
+              : replyType.toLowerCase() as SBUConfig['groupChannel']['channel']['replyType'],
```
bang9, line 122: Is there a safer alternative to type assertion for safer casting? If the `replyType` in React UIKit is refactored into a different form, it seems that there would be no way to detect it using tools like TypeScript.

bang9:
Things to consider for checking:

1. Decide whether to include the `@sendbird/uikit-tools` package in the source code or add it as a dependency
(e.g., `@sendbird/chat` vs `ts-pattern`)

2. We have decided to deprecate the previously used options and migrate them to the newly implemented UIKit configs. Should we consider deprecating React props as well?

src/lib/Sendbird.tsx
AhyoungRyu: How about this way ? 
https://github.com/sendbird/sendbird-uikit-react/pull/614/commits/ef8f1c068361e1da36df5bf018a85c8ba6b28994

bang9 (approved):

PR #616, opened 2023-06-08
feat: apply groupchannel enableMessageSearch config flag

AhyoungRyu:
### Description Of Changes
https://sendbird.atlassian.net/browse/UIKIT-4085

Applied enable_message_search config flag by adding `showSearchIcon` to lib/Sendbird component prop which is not a completely new one but it can be passed from `App` component which already has `showSearchIcon` prop. 

 - [x] group_channel.setting.input.enable_message_search

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -241,6 +241,7 @@ declare module "SendbirdUIKitGlobal" {
     markAsReadScheduler?: MarkAsReadSchedulerType;
     disableUserProfile?: boolean;
     disableMarkAsDelivered?: boolean;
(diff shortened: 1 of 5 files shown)


sravan-s (approved):
LGTM, can you add these props to index_d_ts also 👍

AhyoungRyu:
@sravan-s Oh thanx. I missed that. I added `showSearchIcon` prop to `SendbirdProviderProps` interface https://github.com/sendbird/sendbird-uikit-react/pull/616/commits/d00f4a07589db848f930b72c44235784919091a9

PR #617, opened 2023-06-08
feat: apply groupChannel.enableTypingIndicator config flag

AhyoungRyu:
### Description Of Changes
https://sendbird.atlassian.net/browse/UIKIT-4085

Address enable_typing_indicator configs to ChannelUI component.

- [x] group_channel.channel.channe.enable_typing_indicator

### DIFF
--- a/src/modules/Channel/components/ChannelUI/index.tsx
+++ b/src/modules/Channel/components/ChannelUI/index.tsx
@@ -122,9 +122,10 @@ const ChannelUI: React.FC<ChannelUIProps> = ({
         }
         <div className="sendbird-conversation__footer__typing-indicator">
           {
-            renderTypingIndicator?.() || (
-              <TypingIndicator />
-            )
+            renderTypingIndicator?.()
+              || globalStore?.config?.groupChannel?.enableTypingIndicator && (
+                <TypingIndicator />
+              )
           }
           {
             !isOnline && (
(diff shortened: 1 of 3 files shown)


src/modules/Channel/components/ChannelUI/index.tsx
```diff
@@ -122,9 +122,10 @@ const ChannelUI: React.FC<ChannelUIProps> = ({
         }
         <div className="sendbird-conversation__footer__typing-indicator">
           {
-            renderTypingIndicator?.() || (
-              <TypingIndicator />
-            )
+            renderTypingIndicator?.()
```
AhyoungRyu, line 125: It seems this file is only used for GroupChannel type but let me know if there's any chance that this can be used for OpenChannel too. Will add more condition then.

AhyoungRyu:
I wanted to add a new unit test file for module/ChannelUI but jest throws an error like jest doesn't support ESM. 
Screenshot 2023-06-08 at 12 35 12 PM


It's weird that nothing happened in the files under `ui/*` but it happens only in under `modules/*`. I probably need more time to dig in. Will look into it more separately.

src/modules/Channel/components/ChannelUI/index.tsx
sravan-s: yeah, is in groupChannel only for now ~

sravan-s (approved):

PR #620, opened 2023-06-08
feat: Use channel optionally inside MessageInput

sravan-s:
If customer is using MessageInput inside our modules(ie: Channel, Thread, etc),
we should use the config from the module.
If customer is using MessageInput outside our modules(ie: custom UI),
we expect Channel to be undefined and customer gets control to show/hide file-upload.

### DIFF
--- a/src/ui/MessageInput/messageInputUtils.ts
+++ b/src/ui/MessageInput/messageInputUtils.ts
@@ -2,13 +2,23 @@
  * Write new utils here
  * Migrate old utils as needed, and delete utils.js
  */
-import type { ChannelType } from '@sendbird/chat';
+// import { ChannelType } from '@sendbird/chat';
 import type { GroupChannel } from '@sendbird/chat/groupChannel';
 import type { OpenChannel } from '@sendbird/chat/openChannel';
 import { match } from 'ts-pattern';
 
 import type { SendBirdStateConfig } from '../../lib/types';
 
+/**
+ * FIXME:
+ * Import this ChannelType enum from @sendbird/chat
+ * once MessageInput.spec unit tests can be run \wo jest <-> ESM issue
+ */
+enum ChannelType {
+  BASE = 'base',
+  GROUP = 'group',
+  OPEN = 'open',
+}
 /**
  * FIXME: Simplify this in UIKit@v4
  * If customer is using MessageInput inside our modules(ie: Channel, Thread, etc),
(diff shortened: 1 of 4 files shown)


src/ui/MessageInput/messageInputUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+import { match } from 'ts-pattern';
+
+import type { SendBirdStateConfig } from '../../lib/types';
+
+/**
+ * FIXME: Simplify this in UIKit@v4
+ * If customer is using MessageInput inside our modules(ie: Channel, Thread, etc),
+ * we should use the config from the module.
+ * If customer is using MessageInput outside our modules(ie: custom UI),
+ * we expect Channel to be undefined and customer gets control to show/hide file-upload.
```
sravan-s, line 17: @AhyoungRyu

src/ui/MessageInput/messageInputUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param {*} config SendBirdStateConfig
+ * @returns boolean
+ */
+export const checkIfFileUploadEnabled = ({ channel, config }: {
+  channel?: GroupChannel | OpenChannel,
+  config?: SendBirdStateConfig,
+}) => {
+  const isEnabled = match(channel?.channelType)
+    .with(ChannelType.GROUP, () => config?.openChannel?.enableDocument)
+    .with(ChannelType.OPEN, () => config?.groupChannel?.enableDocument)
```
AhyoungRyu, line 28:
```suggestion
    .with(ChannelType.GROUP, () => config?.groupChannel?.enableDocument)
    .with(ChannelType.OPEN, () => config?.openChannel?.enableDocument)
```

I guess this should be switched?

AhyoungRyu:
Let me fix the failed unit test and merge into the base branch.

src/ui/MessageInput/messageInputUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @returns boolean
+ */
+export const checkIfFileUploadEnabled = ({ channel, config }: {
+  channel?: GroupChannel | OpenChannel,
+  config?: SendBirdStateConfig,
+}) => {
+  const isEnabled = match(channel?.channelType)
+    .with(ChannelType.GROUP, () => config?.groupChannel?.enableDocument)
+    .with(ChannelType.OPEN, () => config?.openChannel?.enableDocument)
+    .otherwise(() => true);
```
AhyoungRyu, line 39: @sravan-s BTW I think there can be a case that user sets the `enableDocument` as `false` from the dashboard but don't pass `channel` prop to this Channel component. 
Then the file icon will still be displayed even the use modified the config value to `false`. What do u think about this scenario?

PR #623, opened 2023-06-08
feat: reset default value of duplicated configs in modules/App

AhyoungRyu:
### Description Of Changes
The current default value setting via prop-types makes every configs' value have default value from the App component level. This makes Sendbird config priotization logic unnecessary. 

So I reset configs' value to `undefined` to make them set in Sendbird component instead.

### DIFF
--- a/src/modules/App/index.jsx
+++ b/src/modules/App/index.jsx
@@ -163,7 +163,6 @@ App.defaultProps = {
   allowProfileEdit: false,
   onProfileEditSuccess: null,
   disableMarkAsDelivered: false,
-  showSearchIcon: false,
   renderUserProfile: null,
   config: {},
   voiceRecord: {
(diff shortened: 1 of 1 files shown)


AhyoungRyu:
@sravan-s I closed the App.jsx TS migration PR #621 but still need this logic to make the config prioritization logic work in Sendbird component > UIKitConfigProvider.

sravan-s:
This would change UX for existing customer
Lets make / export a new component?

AhyoungRyu:
>This would change UX for existing customer

In terms of how the UX can be changed? Just curious 😄

sravan-s (approved):

AhyoungRyu:
Let me leave a comment where we initialize those configs' default values if a customer doesn;t specify the config prop values when they use this modules/App component.

**AS-IS**
1. modules/App component in `App.defaultProps` by `props-types`.
2. Sendbird component which is called by modules/App 
So basically we used to have two different initialize fallback logic in 1 & 2.

**TO-BE**
Sendbird component only. It contains UIKitConfiguration inside and this will be in charge of the config value initailization. 


The each config's default value is mapped in this way. They have same default values. So basically this change should not impact on the existing behavior.  
```
  showSearchIcon = groupChannel.setting.enableMessageSearch: false.
  isMentionEnabled = groupChannel.channel.enableMention: false,
  isReactionEnabled =  groupChannel.channel.enableMention: true,
  replyType groupChannel.channel.replyType: 'QUOTE_REPLY',
  disableUserProfile = !common.enableUsingDefaultUserProfile: false,
  isVoiceMessageEnabled = groupChannel.channel.enableVoiceMessage: false,
  isTypingIndicatorEnabledOnChannelList = groupChannel.channelList.enableTypingIndicator: false,
  isMessageReceiptStatusEnabledOnChannelList = groupChannel.channelList.enableMessageReceiptStatus: false,
```
The new config init logic is in [here](https://github.com/sendbird/sendbird-uikit-core-ts/blob/main/packages/uikit-tools/src/uikitConfiguration/internal/initialConfig.ts#L4).

sravan-s:
Just use this as commit message ㅋㅋㅋㅋ 👍

PR #624, opened 2023-06-08
[Message]: Make the message component selectable

jawakarD:
I know I'm not supposed to open PR as I'm not a contributer, but please take a look and let me know. I recently upgraded and found this. If I'm removing something that's essential, let me know I can try to help. I found this code is added from #550, so if the code I'm removing is supposed to work only on mobile, I'll try to add a media-query. Won't be offended if you close the PR also :)

## External Contributions

and we'll evaluate if we can setup a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) to allow for the contribution.

### Description Of Changes

- Not able to select messages even email addresses from the message.

https://github.com/sendbird/sendbird-uikit-react/assets/31589372/b4346f7e-445d-4e6c-b455-0e921356492d

### DIFF
--- a/src/ui/Label/index.scss
+++ b/src/ui/Label/index.scss
@@ -1,10 +1,5 @@
 @import '../../styles/variables';
 
-.sendbird-label {
-  -webkit-user-select: none;
-  -webkit-touch-callout: none;
-}
-
 [class*=sendbird-label] {
   font-family: var(--sendbird-font-family-default);
 }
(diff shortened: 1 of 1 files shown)


src/ui/Label/index.scss
```diff
@@ -1,10 +1,5 @@
 @import '../../styles/variables';
 
-.sendbird-label {
-  -webkit-user-select: none;
-  -webkit-touch-callout: none;
-}
```
sravan-s, line 6: Hey, thanks for this, but this is a workaround for certain mobile devices... anyways, will scope it down so that it works on only those devices ~

sravan-s: Also, sadly, we are not allowed to merge these PRs from ext. I will make a  separate PR and add you as courtesy

sravan-s: https://github.com/sendbird/sendbird-uikit-react/pull/625

PR #631, opened 2023-06-09
fix: Thread / Add extra state to show/hide delete menu

sravan-s:
message?.threadInfo?.replyCount is not getting updated dynamically

This is a temporary prop that we wont open to public, if this is SDK issue,
then we can remove later 

Fixes: https://sendbird.atlassian.net/browse/UIKIT-4098

### DIFF
--- a/src/modules/Thread/components/ParentMessageInfo/index.tsx
+++ b/src/modules/Thread/components/ParentMessageInfo/index.tsx
@@ -339,6 +339,11 @@ export default function ParentMessageInfo({
           hideMenu={() => {
             setShowMobileMenu(false);
           }}
+          deleteMenuState={
+            allThreadMessages?.length === 0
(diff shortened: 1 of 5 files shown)


src/modules/Thread/components/ParentMessageInfo/index.tsx
```diff
@@ -339,6 +339,11 @@ export default function ParentMessageInfo({
           hideMenu={() => {
             setShowMobileMenu(false);
           }}
+          deleteMenuState={
+            allThreadMessages?.length === 0
+              ? 'ACTIVE'
+              : 'HIDE'
+          }
```
AhyoungRyu, line 346: Q. what does it mean by this? thread message's length is 0 then Active if not, hide?

AhyoungRyu (approved):
Just left a small question.

PR #634, opened 2023-06-12
chore: Add sample to reroute when customer is banned

sravan-s:
fixes: https://sendbird.atlassian.net/browse/QM-2172

### DIFF
--- a/samples/router/src/App.tsx
+++ b/samples/router/src/App.tsx
@@ -8,18 +8,16 @@ import { router as desktopRouter } from './DesktopLayout/Router'
 function App() {
   const isMobile = /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
   return (
-    <>
-      <div className="sendbird-app">
-        <SendbirdProvider
-          appId="YOUR_APP_ID"
-          userId='YOUR_USER_ID'
-          nickname='YOUR_NICKNAME'
-          breakpoint={isMobile}
-        >
-          <RouterProvider router={isMobile ? mobileRouter : desktopRouter} />
-        </SendbirdProvider>
-      </div>
-    </>
+    <div className="sendbird-app">
+      <SendbirdProvider
+        appId="YOUR_APP_ID"
+        userId='YOUR_USER_ID'
+        nickname='YOUR_NICKNAME'
+        breakpoint={isMobile}
+      >
+        <RouterProvider router={isMobile ? mobileRouter : desktopRouter} />
+      </SendbirdProvider>
+    </div>
   )
 }
 
--- a/samples/router/src/App.tsx
+++ b/samples/router/src/App.tsx
@@ -1,48 +1,25 @@
-import '@sendbird/uikit-react/dist/index.css'
+import './App.css'
 
 import SendbirdProvider from '@sendbird/uikit-react/SendbirdProvider'
-import {
-  createBrowserRouter,
-  RouterProvider,
-} from 'react-router-dom'
-
-import { ErrorPage } from './routes/Error'
-import { ChannelList } from './routes/ChannelList'
-import { Channel } from './routes/Channel'
-import { ChannelSettings } from './routes/ChannelSettings'
-import { MessageSearch } from './routes/MessageSearch'
-
-const router = createBrowserRouter([
-  {
-    path: '/',
-    element: <ChannelList />,
-    errorElement: <ErrorPage />,
-  },
-  {
-    path: 'channel/:channelUrl',
-    element: <Channel />,
-  },
-  {
-    path: 'channel/:channelUrl/settings',
-    element: <ChannelSettings />,
-  },
-  {
-    path: 'channel/:channelUrl/search',
-    element: <MessageSearch />,
-  },
-])
+import { RouterProvider } from 'react-router-dom'
+import { router as mobileRouter } from './MobileLayout/Router'
+import { router as desktopRouter } from './DesktopLayout/Router'
 
 function App() {
+  const isMobile = /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)
   return (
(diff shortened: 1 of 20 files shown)


samples/router/src/hooks/useCustomRouting.ts
```diff
@@ -0,0 +1,36 @@
...
+  const navigate = useNavigate()
+  const sb = store?.stores?.sdkStore?.sdk
+  const userId = store?.config?.userId
+  useEffect(() => {
+    const handlerId = simpleUUID()
+    if (sb?.groupChannel?.addGroupChannelHandler) {
+      sb.groupChannel.removeGroupChannelHandler(handlerId)
+      const handler = new GroupChannelHandler({
+        onUserBanned: (groupChannel, user) => {
+          if (groupChannel.url === currentChannel?.url && user?.userId === userId) {
```
AhyoungRyu, line 22: It seems you passed `channelUrl` from `routes/Channel` component 
https://github.com/sendbird/sendbird-uikit-react/pull/634/files#diff-cebbd30c8ec7f1be5cdc180fd2d6c315d8caa9d0830428bad7d6dcadc77cc2f2R12 but retrieving `url` again?

sravan-s: ah yeah :D

samples/router/src/hooks/useCustomRouting.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect } from "react"
+import useSendbirdStateContext from "@sendbird/uikit-react/useSendbirdStateContext"
+import { GroupChannelHandler } from '@sendbird/chat/groupChannel'
+import { useNavigate } from 'react-router-dom'
+
+
+const simpleUUID = () => {
+  return Date.now().toString(36) + Math.random().toString(36).substr(2)
+}
```
HoonBaek, line 9: nit: we have an `uuid` util function

samples/router/src/routes/Channel.tsx
```diff
@@ -8,6 +9,7 @@ export function Channel() {
   const createdAt = searchParams.get('createdAt')
   const numCreatedAt = Number(createdAt)
   const navigate = useNavigate()
+  useCustomRouting({ currentChannel: channelUrl })
```
HoonBaek, line 12: Does this `channelUrl` here actually mean `channel`?
The `currentChannel` should be a channel including the channel URL, instead of only URL

samples/router/src/hooks/useCustomRouting.ts
sravan-s: We cannot really use this for the sample ~ customers  should be able to use this as "self contained"

HoonBaek: Ah this is to show how to implement it, I got it

sravan-s (commented):
I made the code with some `duplication` on purpose ~ Duplication as in -> Desktop/Channel & Mobile/Channel are almost the same component

The purpose is that, users can concentrate more on logic of routing and lay-outing instead of what's going on inside each individual components

samples/router/src/App.tsx
```diff
@@ -1,48 +1,25 @@
...
-    <div className="sendbird-app">
-      <SendbirdProvider
-        appId={import.meta.env.VITE_SB_APP_ID}
-        userId='sendbird'
-        nickname='sendbird'
-      >
-        <RouterProvider router={router} />
-      </SendbirdProvider>
-    </div>
+    <>
```
HoonBaek: nit: we can remove this

samples/router/src/DesktopLayout/Channel.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export default function Channel() {
+  const { channelUrl } = useParams()
+  const [searchParams] = useSearchParams()
+  const messageId = searchParams.get('messageId')
+  const createdAt = searchParams.get('createdAt')
+  const numCreatedAt = Number(createdAt)
+  const navigate = useNavigate()
+  return (
+    <>
```
HoonBaek: nit: good to remove this too

samples/router/src/DesktopLayout/Channel.tsx
```diff
@@ -0,0 +1,42 @@
...
+            navigate('/')
+          }}
+          onSearchClick={() => {
+            navigate(`/channels/${channelUrl}/search`)
+          }}
+          { ...(messageId && typeof numCreatedAt === 'number') && {
+            startingPoint: numCreatedAt,
+            highlightedMessageId: messageId,
+            animatedMessageId: messageId,
+          }}
```
HoonBaek: This operation could return `false`, no?

samples/router/src/MobileLayout/ChannelSettings.tsx
```diff
@@ -1,16 +1,18 @@
...
 
+import { useNavigateOnBan } from '../hooks/useNavigateOnBan'
+
 export function ChannelSettings() {
   const { channelUrl } = useParams()
   const navigate = useNavigate()
+  useNavigateOnBan()
   return (
-    <div>
+    <div style={{ height: '100vh' }}>
```
HoonBaek, line 11: Can we use `100%` instead of using `100vh`? Because it breaks some CSS.

HoonBaek (approved):

samples/router/src/MobileLayout/ChannelSettings.tsx
sravan-s: 100% is probably cleaner, but, this is sample, doesnt matter too much

samples/router/src/DesktopLayout/Channel.tsx
sravan-s: yeah, but its okay :)

PR #639, opened 2023-06-12
fix: use global disableUserProfile if each context's one is defined

AhyoungRyu:
## External Contributions
Resolves https://sendbird.atlassian.net/browse/UIKIT-4139

If any provider which uses `UserProfileContext` doesn't have a prop value for `disableUserProfile`  but only global config has the value, the latter is not reflected properly. 
So I put a fallback value so the global config can be used. If both are undefined, then the `disableUserProfile` will be set as undefined.

### DIFF
--- a/src/modules/Channel/context/ChannelProvider.tsx
+++ b/src/modules/Channel/context/ChannelProvider.tsx
@@ -485,7 +485,7 @@ const ChannelProvider: React.FC<ChannelContextProps> = (props: ChannelContextPro
       toggleReaction,
     }}>
       <UserProfileProvider
-        disableUserProfile={props?.disableUserProfile}
+        disableUserProfile={props?.disableUserProfile ?? config?.disableUserProfile}
         renderUserProfile={props?.renderUserProfile}
         onUserProfileMessage={onUserProfileMessage}
       >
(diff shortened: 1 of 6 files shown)


HoonBaek (commented):
`props?.disableUserProfile`의 기본값을 넣어주는 곳이 있는지 확인이 필요할 것 같습니다. 기본값이 세팅된다면 `??` operation이 무용할 것 같아서요

I'm curious if there's a chance that the variable becomes null. I guess it's set from somewhere, so could you confirm it?

sravan-s (approved):

AhyoungRyu:
@HoonBaek 
Screenshot 2023-06-13 at 10 22 54 AM
Screenshot 2023-06-13 at 10 25 27 AM
Screenshot 2023-06-13 at 10 25 40 AM

I guess there's no place to set the prop as null at the moment? I reset this prop value in modules/App/index.js in [this PR](https://github.com/sendbird/sendbird-uikit-react/pull/623/files#diff-6e8d0a17bdb300f8a4dcb5cc0a7f10e6963dd4a6125aaab9e653b0fb138004a3R185) and lib/Sendbird too so that the config value can be applied \w Dashboard setting.

But even before, its default value was true / false / undefined I think.

PR #643, opened 2023-06-14
fix: Handle undefined message type in is voice message

HoonBaek:
### Description Of Changes

issue: Screen error shows up intermittently and says that the `message.type` property is undefined
fix: Check if the type property is empty and return false when it is

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -165,10 +165,10 @@ export const isAudioMessage = (message: FileMessage): boolean => message && isFi
 export const isAudioMessageMimeType = (type: string): boolean => (/^audio\//.test(type));
 export const isVoiceMessage = (message: Nullable<FileMessage | UserMessage>): boolean => {
   // ex) audio/m4a OR audio/m4a;sbu_type=voice
-  if (!(message && isFileMessage(message))) {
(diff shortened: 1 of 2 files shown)


src/utils/index.ts
```diff
@@ -163,23 +163,25 @@ export const isGifMessage = (message: UserMessage | FileMessage): boolean => (
...
+  const [mimeType, typeParameter] = (message as FileMessage).type.split(';');
+
+  if (!isAudioMessageMimeType(mimeType)) {
     return false;
   }
-  const [key, value] = typeParameter.split('=');
-  if (key === 'sbu_type' && value === 'voice') {
-    return true;
+
+  if (typeParameter) {
```
AhyoungRyu, line 178: Do we want to pass the case `typeParameter === undefined` to [the last return statement](https://github.com/sendbird/sendbird-uikit-react/pull/643/files#diff-6147e3c1761df71fd40952dbcbac388a45f80b8c318f367ef41048351a2c0c99R183)?

HoonBaek: Yes I do

AhyoungRyu (approved):

HoonBaek:
https://sendbird.atlassian.net/browse/UIKIT-4144

PR #647, opened 2023-06-19
chore: Remove react-dom/server from rollup.config

sravan-s:
If a path is set on Rollup.config.externals, it wont be bundled
to the library during rollup build step. We have removed react-dom/server
in: https://github.com/sendbird/sendbird-uikit-react/pull/449
So, its okay to remove it

Fixes: https://sendbird.atlassian.net/browse/UIKIT-4191


sravan-s:
Build works:
Screen Shot 2023-06-19 at 2 45 33 PM

AhyoungRyu (approved):

PR #648, opened 2023-06-19
fix: Race condition in disconnect/connect

sravan-s:
When there is only single instance of SDK,
and disconnect call take longer than next connect call
Solution: Set `newInstance: true` when initializing SDK instance

Fixes: https://sendbird.atlassian.net/browse/UIKIT-4193


AhyoungRyu (approved):
Thanks!

PR #649, opened 2023-06-20
fix: Allow to reduce mobile app height

HoonBaek:
### Description Of Changes

issue: It was not able to reduce the height of the mobile app with some wrapper component

* fix: Set the height of MobileApp to 100% & do not use 'vh'

[UIKIT-4188](https://sendbird.atlassian.net/browse/UIKIT-4188)

[UIKIT-4188]: https://sendbird.atlassian.net/browse/UIKIT-4188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/App/stories/utils.jsx
+++ b/src/modules/App/stories/utils.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
 
-export const fitPageSize = (component) => (
+export const fitPageSize = (component, style = {}) => (
   <div
     style={{
       height: '100vh',
       width: '100vw',
+      ...style,
     }}
   >
     {component}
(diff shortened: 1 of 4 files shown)


AhyoungRyu (approved):

src/modules/App/stories/utils.jsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 
-export const fitPageSize = (component) => (
+export const fitPageSize = (component, style = {}) => (
```
sravan-s, line 3: do we use styles somewhere and is not applied? curious

sravan-s (approved):

src/modules/App/stories/utils.jsx
HoonBaek: no, I added this just because it's useful when we test it in local

PR #658, opened 2023-06-27
feat: add gh workflow for auto release ticket creation

AhyoungRyu:
### Description Of Changes
Addresses https://sendbird.atlassian.net/browse/CLNP-176 
This PR is a very beginning the project of automating our release process. Two small things are done in this PR, 

- [x] Added a new gh workflow file to receive a signal from outside of Github  e.g. Slack / GIthub CLI / Other 3party tools 
  👉🏻 This enables to create a release Jira ticket without navigating to Jira board or Github. 
(Got an idea from [a proposal](https://sendbird.atlassian.net/wiki/spaces/UIKitreact/pages/2101838231/Release+Automation#Concrete-Proposal-(June-23---2023)) from @sravan-s, see the Slack bot section)
Once this workflow file is merged into `main` branch, running this below command from a Slack channel will create a ticket.  Note that the command accepts only `branch_name` at the moment. 
```
/github workflow dispatch sendbird/sendbird-uikit-react --workflow release-ticket-reaction.yml -i branch_name=your-branch-name
```
The newly created ticket can be found in #sdk-release-approver channel in Slack. You can ask the approvers to check the ticket in the channel.
But yeah, the cmd is too long to remember. So I'm thinking of creating a short custom command in Slack for the sake of easy-use.

- [x] Modified the Circle CI config to take an advantage of existing Jira ticket creation pipeline 
 👉🏻 got the most of logic from the chat ios repo; https://github.com/sendbird/chat-ios/blob/develop/.github/workflows/pr-comment-bot.yml#L90

PR #660, opened 2023-06-27
fix: incorrect Circle CI secret env name

AhyoungRyu:
This is a following PR after #658 merged.  `CIRCLECI_API_TOKEN` is the right name not `CIRCLE_CI_API_TOKEN`.

### DIFF
--- a/.github/workflows/release-ticket-creation.yml
+++ b/.github/workflows/release-ticket-creation.yml
@@ -3,33 +3,36 @@ on:
   workflow_dispatch:
     inputs:
       branch_name:
-        description: 'Release target branch name'
+        description: 'Type a branch name starting with `release/v`'
         required: true
 
 jobs:
   trigger-release-ticket-creation:
     name: Trigger release ticket creation
     runs-on: ubuntu-latest
-    # The branch name should starts with 'release/v'
-    if: startsWith(github.event.inputs.branch_name, 'release/v')
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+
+      - name: Verify branch name
+        run: |
+          if [[ ! ${{ github.event.inputs.branch_name }} =~ ^release/v ]]; then
+            echo "Branch name should start with 'release/v'"
+            exit 1
+          fi
+
       - name: Trigger CircleCI Job
         run: |
           API_RESULT=$(curl --request POST \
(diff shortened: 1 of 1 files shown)


.github/workflows/release-ticket-creation.yml
```diff
@@ -3,33 +3,36 @@ on:
   workflow_dispatch:
     inputs:
       branch_name:
-        description: 'Release target branch name'
+        description: 'Type a branch name starting with `release/v`'
```
AhyoungRyu, line 6: This field is displayed in the workflow trigger popup UI like this. 
So I aded a short guide message to make the us type `release/v` branch name. (Too bad that I can set `placeholder` 🫠

Screenshot 2023-06-27 at 6 02 48 PM

.github/workflows/release-ticket-creation.yml
```diff
@@ -3,33 +3,36 @@ on:
...
   trigger-release-ticket-creation:
     name: Trigger release ticket creation
     runs-on: ubuntu-latest
-    # The branch name should starts with 'release/v'
-    if: startsWith(github.event.inputs.branch_name, 'release/v')
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+
+      - name: Verify branch name
```
AhyoungRyu, line 17: Why I changed the one line cmd like this? 
With 
```
if: startsWith(github.event.inputs.branch_name, 'release/v')
```
the workflow will skipped in silence but I want to make it failed \w the error message so we know why it failed.

.github/workflows/release-ticket-creation.yml
```diff
@@ -3,33 +3,36 @@ on:
...
               "branch": "${{ github.event.inputs.branch_name }}",
               "parameters": {
-                "run_workflow_create_ticket": true,
+                "run_workflow_create_ticket": true
               }
             }')
           echo "API_RESULT: ${API_RESULT}"
           CIRCLE_CI_JOB_NUMBER=$(echo "${API_RESULT}" | jq -r '.number')
           echo "::set-output name=DEPLOY_COMMENT_BODY::https://app.circleci.com/pipelines/github/${{ github.repository }}/$CIRCLE_CI_JOB_NUMBER"
-        env:
```
AhyoungRyu, line 33: Don't need this part.

HoonBaek (approved):
LGTM

PR #665, opened 2023-06-30
feat: enable channel creation for logged-in user only

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-251

Enabled the channel creation button if there's only the logged in user && users.length === 0.

### DIFF
--- a/src/modules/CreateChannel/components/InviteUsers/index.tsx
+++ b/src/modules/CreateChannel/components/InviteUsers/index.tsx
@@ -91,10 +91,12 @@ const InviteUsers: React.FC<InviteUsersProps> = ({
       titleText={titleText}
       submitText={submitText}
       type={ButtonTypes.PRIMARY}
-      disabled={Object.keys(selectedUsers).length === 0}
+      disabled={users.length > 0 && Object.keys(selectedUsers).length === 0}
       onCancel={onCancel}
       onSubmit={() => {
-        const selectedUserList = Object.keys(selectedUsers);
+        const selectedUserList = Object.keys(selectedUsers).length > 0
+          ? Object.keys(selectedUsers)
+          : [userId];
         if (typeof overrideInviteUser === 'function') {
           overrideInviteUser({
             users: selectedUserList,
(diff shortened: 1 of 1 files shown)


src/modules/CreateChannel/components/InviteUsers/index.tsx
```diff
@@ -103,28 +105,27 @@ const InviteUsers: React.FC<InviteUsersProps> = ({
           });
           return;
         }
-        if (selectedUserList.length > 0) {
```
AhyoungRyu, line 106: Nothing changed in this if block, but just removed the condition cause there'll be no case this `selectedUserList.length === 0` with [this logic](https://github.com/sendbird/sendbird-uikit-react/pull/665/files#diff-08c2d4af32c32fa2cabfaf2e14278e24dea1f367e411a2483e84f38bdb73790cR97)

src/modules/CreateChannel/components/InviteUsers/index.tsx
```diff
@@ -91,10 +91,12 @@ const InviteUsers: React.FC<InviteUsersProps> = ({
       titleText={titleText}
       submitText={submitText}
       type={ButtonTypes.PRIMARY}
-      disabled={Object.keys(selectedUsers).length === 0}
+      disabled={users.length > 0 && Object.keys(selectedUsers).length === 0}
```
sravan-s, line 94: If possible, can you add a  test for this -> 
`should enable create button when there are no users in the list` 
`should disable create button when there are users on list but none's checked` 

Separate PR is okay

Else, comments are fine too

sravan-s (approved):

src/modules/CreateChannel/components/InviteUsers/index.tsx
sravan-s: Because this is a bit unobvious part ~

HoonBaek, line 94: This logic stil disable the submit button when there's no selected user

HoonBaek (approved):

src/modules/CreateChannel/components/InviteUsers/index.tsx
AhyoungRyu: Right. This part is related to this conversation https://sendbird.slack.com/archives/C050Z6DA70A/p1688089591024089?thread_ts=1688086902.570209&cid=C050Z6DA70A if I got your point correctly

AhyoungRyu: https://github.com/sendbird/sendbird-uikit-react/pull/665/commits/16b47590799c020590999022f86408a7a652aa9b added a comment first. Will add some unit tests separately!

PR #667, opened 2023-06-30
feat: make jira ticket creation workflow fail explicitly

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-255

Currently, the release Jira ticket creation workflow gets passed even though there's an error with Circle API request. 
So made a change to get failed when it gets an error on this step. 
Screenshot 2023-06-30 at 2 54 08 PM

With this change, the warning in the image should be gone too. Just followed a suggestion here https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/release-ticket-creation.yml
+++ b/.github/workflows/release-ticket-creation.yml
@@ -21,6 +21,26 @@ jobs:
             exit 1
           fi
 
+      - name: Check if branch exists
+        id: check_branch
+        run: |
+          set -x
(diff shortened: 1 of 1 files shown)


AhyoungRyu:
Let me test a bit more

sravan-s (approved):

AhyoungRyu:
Will close this for now since we're not using this script ATM.

PR #674, opened 2023-07-06
fix: Move message list scroll when the last message is edited

HoonBaek:
* remove the duplicated functions: handleMessageHeightChange, and handleMessageListHeightChange
  AND combine them into one function: `moveScroll`

* add optional params to the `moveScroll`
  for moving the scroll only when the last message is reached to the bottom

* move scroll only when the last message's updatedAt is changed

### DIFF
--- a/src/modules/Channel/components/Message/index.tsx
+++ b/src/modules/Channel/components/Message/index.tsx
@@ -33,8 +33,7 @@ type MessageUIProps = {
   hasSeparator?: boolean;
   chainTop?: boolean;
   chainBottom?: boolean;
-  handleScroll?: () => void;
-  handleMessageListHeightChange?: () => void;
+  handleScroll?: (isAffectedOnlyGround?: boolean) => void;
   // for extending
   renderMessage?: (props: RenderMessageProps) => React.ReactElement;
   renderCustomSeparator?: (props: RenderCustomSeparatorProps) => React.ReactElement;
@@ -49,7 +48,6 @@ const Message = ({
   chainTop,
   chainBottom,
   handleScroll,
-  handleMessageListHeightChange,
   renderCustomSeparator,
   renderEditInput,
   renderMessage,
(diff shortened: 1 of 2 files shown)


src/modules/Channel/components/Message/index.tsx
```diff
@@ -33,8 +33,7 @@ type MessageUIProps = {
   hasSeparator?: boolean;
   chainTop?: boolean;
   chainBottom?: boolean;
-  handleScroll?: () => void;
-  handleMessageListHeightChange?: () => void;
+  handleScroll?: (isAffectedOnlyGround?: boolean) => void;
```
sravan-s, line 36: isAffectedOnlyGround is not really clear, lets improve the name?

HoonBaek: how about `isBottomMessageAffected`? is it more clear? or any suggestion?

sravan-s: both sounds okay

sravan-s (approved):
@AhyoungRyu , can you also check?

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -120,29 +120,22 @@ const MessageList: React.FC<MessageListProps> = ({
...
   /**
-   * Move the messsage list scroll
-   * when each message's height is changed by `reactions` OR `showEdit`
+   * 1. Move the messsage list scroll
+   *    when each message's height is changed by `reactions` OR `showEdit`
+   * 2. Keep the scrollBottom value after fetching new message list
    */
-  const handleMessageHeightChange = () => {
+  const moveScroll = (_isBottomMessageAffected?: boolean): void => {
+    const isBottomMessageAffected = _isBottomMessageAffected ?? false;
```
AhyoungRyu, line 128:
```suggestion
  const moveScroll = (isBottomMessageAffected = false): void => {
```
This should be fine too :)

AhyoungRyu (approved):

src/modules/Channel/components/MessageList/index.tsx
HoonBaek: aha! thank you~

AhyoungRyu:
@HoonBaek Thanks for your help, Hoon!

PR #675, opened 2023-07-06
fix: Update webpack to 5.88.1

sravan-s:
To make `Cross-realm object access in Webpack 5` go away
See: https://nvd.nist.gov/vuln/detail/CVE-2023-28154

Webpack is somewhere down below in dependency list,
so just update package-lock ~

Bonus: Remove APP_ID from consts

Fixes: https://sendbird.atlassian.net/browse/CLNP-308

### DIFF
--- a/samples/typescript_sample/package-lock.json
+++ b/samples/typescript_sample/package-lock.json
@@ -24,73 +24,81 @@
         "web-vitals": "^2.1.4"
       }
     },
+    "node_modules/@aashutoshrathi/word-wrap": {
+      "version": "1.2.6",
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
Wouldn't it be override the changed package-lock if we run `npm install` under the TS sample directory?

sravan-s:
Hmmm, if you install something new,  or use old version of npm, yeah.. Otherwise no.. But, we really dont need to worry about this.

We can even just delete that file

This is to make this guy silent
Screen Shot 2023-07-06 at 10 48 25 AM

AhyoungRyu (approved):
Yeah looks good as a workaround

PR #679, opened 2023-07-06
fix: scroll to bottom on message.data.stream change

AhyoungRyu:
Unfortunately when AI Chatbot is sending a streaming message, `updatedAt` field doesn't change but only `message.data.stream: true / false` changed in the meantime. 
So added some additional logic to the part made by #674 accordingly.


HoonBaek (approved):
It could scroll once at the last when the Chat is finishing chat.

AhyoungRyu:
@HoonBaek Yeah right. I asked the server engineer to change the `updatedAt`. I might close this PR if they address the change quickly. Let me wait a bit.

PR #687, opened 2023-07-19
feat: create separate package.json for cjs module on build time

AhyoungRyu:
Addresses https://sendbird.slack.com/archives/G01290GCDCN/p1689744629102179?thread_ts=1689719295.903759&cid=G01290GCDCN

Screenshot 2023-07-19 at 3 16 12 PM

PR #683, opened 2023-07-12
feat: Add a new props isUserIdUsedForNickname

HoonBaek:
* feat: Add isUserIdUsedForNickname to the public interface
* feat: Apply isUserIdUsedForNickname into the connection logic
* feat: Set userId to nickname when isUserIdUsedForNickname is true

[UIKIT-2537](https://sendbird.atlassian.net/browse/UIKIT-2537)

[UIKIT-2537]: https://sendbird.atlassian.net/browse/UIKIT-2537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #689, opened 2023-07-19
release: 3.6.4

AhyoungRyu:
### Feat:
* Create a separate package.json for CommonJS (cjs) module during build time. This package.json is located under dist/cjs directory. (#687)
* Add a new prop `isUserIdUsedForNickname` to the public interface. This prop allows using the userId as the nickname. (#683)
* Add an option to the ChannelProvider: `reconnectOnIdle`(default: true), which prevents data refresh in the background. (#690)

### Fixes:
* Fix an issue where the server returns 32 messages even when requesting 31 messages in the Channel. Now, hasMorePrev will not be set to false when the result size is larger than the query. (#688)
* Verify the fetched message list size with the requested size of the MessageListParams. Added a test case for verifying the fetched message list size. (#686)
* Address the incorrect cjs path in package.json. The common js module path in the pacakge.json has been fixed. (#685)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog - v3
 
+## [v3.6.4] (July 19 2023)
+### Feat:
+* Created a separate package.json for CommonJS (cjs) module during build time. This package.json is located under dist/cjs directory. (#687)
+* Added a new prop `isUserIdUsedForNickname` to the public interface. This prop allows using the userId as the nickname. (#683)
+
+### Fixes:
+* Fixed an issue where the server returns 32 messages even when requesting 31 messages in the Channel. Now, hasMorePrev will not be set to false when the result size is larger than the query. (#688)
+* Verified the fetched message list size with the requested size of the MessageListParams. Added a test case for verifying the fetched message list size. (#686)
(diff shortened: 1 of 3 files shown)


AhyoungRyu:
I'll rebase again if there's more commit to include.

sendbird-sdk-deployment:
[Creating Ticket] 🔖 Done https://sendbird.atlassian.net/browse/SDKRLSD-913

sravan-s (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog - v3
 
+## [v3.6.4] (July 19 2023)
+### Feat:
+* Created a separate package.json for CommonJS (cjs) module during build time. This package.json is located under dist/cjs directory. (#687)
```
HoonBaek, line 5: Can we use the present tense for consistency?

AhyoungRyu: Yeah updated it https://github.com/sendbird/sendbird-uikit-react/pull/689/commits/612d30e9b25ce77d76162cfef6a832ad30d284f2

AhyoungRyu:
I added one more commit #690

PR #690, opened 2023-07-19
feat: add `reconnectOnIdle` option to ChannelProvider

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-200 

Some users might not want to refetch the message list when the active browser tab is in the background and then comes back to the foreground. 
So, I added an option to the ChannelProvider: `reconnectOnIdle` (default: true), which prevents data refresh in the background.


AhyoungRyu:
Trying to add more test cases but ready for review for the main logic.

sravan-s (approved):

HoonBaek (approved):

PR #692, opened 2023-07-21
feat: add sdkInitParams prop

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-217

Currently, there's no way to customize `sdk.init()` parameters outside of UIKit since sdk.init() is called internally within UIKit.
So I'm going to add a new prop called sdkInitParams that allows passing custom parameters from outside of UIKit(e.g. chat-ai-widget).

Then user will be able to pass any `SendbirdChatParams` listed in https://sendbird.com/docs/chat/v4/javascript/ref/interfaces/_sendbird_chat.SendbirdChatParams.html via `sdkInitParams`.

### DIFF
--- a/src/lib/hooks/useConnect/setupConnection.ts
+++ b/src/lib/hooks/useConnect/setupConnection.ts
@@ -8,6 +8,7 @@ import { USER_ACTIONS } from '../../dux/user/actionTypes';
 import { isTextuallyNull } from '../../../utils';
 
 import { SetupConnectionTypes } from './types';
+import { SendbirdChatInitParams } from '../../types';
 
 const APP_VERSION_STRING = '__react_dev_mode__';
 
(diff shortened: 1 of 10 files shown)


sravan-s (approved):

sravan-s (approved):
Lets add example to commit mesage 👍

src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -37,6 +40,7 @@ export function setUpParams({
       new OpenChannelModule(),
     ],
     newInstance: true,
+    ...sdkInitParams,
```
sravan-s, line 43: undefined/null check necessary?

AhyoungRyu: Trying spread from undefined happens nothing actually but always better to be explicit :) a2f0541863b6f4ef5460af4a3792d893890f16cf

sravan-s: hmmm have crashed on me once before(dont remember the env)

PR #693, opened 2023-07-21
feat: omit 'appId' from SendbirdChatInitParams

AhyoungRyu:
Addressing a feedback from @sravan-s after #692 merged. `appId` is better not to override just to be safe.

### DIFF
--- a/src/lib/hooks/useConnect/setupConnection.ts
+++ b/src/lib/hooks/useConnect/setupConnection.ts
@@ -34,13 +34,14 @@ export function setUpParams({
   sdkInitParams?: SendbirdChatInitParams;
 }): SendbirdChat {
   const params = {
-    appId,
     modules: [
       new GroupChannelModule(),
       new OpenChannelModule(),
     ],
     newInstance: true,
-    ...sdkInitParams,
+    ...(sdkInitParams ?? {}),
+    // appId shouldn't be overrided
+    appId,
   };
   if (customApiHost) {
     params['customApiHost'] = customApiHost;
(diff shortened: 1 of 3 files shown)


src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -34,13 +34,14 @@ export function setUpParams({
...
 }): SendbirdChat {
   const params = {
-    appId,
     modules: [
       new GroupChannelModule(),
       new OpenChannelModule(),
     ],
     newInstance: true,
-    ...sdkInitParams,
+    ...(sdkInitParams ?? {}),
```
AhyoungRyu, line 42: @sravan-s So weird that 
Also, https://github.com/sendbird/sendbird-uikit-react/commit/a2f0541863b6f4ef5460af4a3792d893890f16cf  this commit from #692 is gone. So had to change this again 🤷‍♂️

src/lib/types.ts
```diff
@@ -247,4 +247,4 @@ export type UIKitOptions = PartialDeep<{
   openChannel: SBUConfig['openChannel']['channel'];
 }>;
 
-export type SendbirdChatInitParams = SendbirdChatParams<Module[]>;
+export type SendbirdChatInitParams = Omit<SendbirdChatParams<Module[]>, 'appId'>;
```
sravan-s, line 250: 👍

sravan-s: should be same in https://github.com/sendbird/sendbird-uikit-react/blob/main/scripts/index_d_ts#L116

src/lib/hooks/useConnect/setupConnection.ts
sravan-s: yeah weird

sravan-s (approved):

PR #682, opened 2023-07-12
feat: Add Sendbird Extensions in connection

HoonBaek:
* feat: Add Sendbird Extensions in connection

[CLNP-197](https://sendbird.atlassian.net/browse/CLNP-197)

[CLNP-197]: https://sendbird.atlassian.net/browse/CLNP-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #698, opened 2023-08-03
feat: add customExtensionParams for sdk.addSendbirdExtensions

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-264

chat-ai-widget is internally using UIKit React and it also has a requirement for tracking a user agent data via the logic added by #682. 

But [the 3rd parameter in the `sdk.addSendbirdExtension`; customData](https://github.com/sendbird/chat-js/pull/1104/files#diff-b1f1de7e79042a6f46acdd47d27f64e33177077d9a28b7a8740cfdbf0c60266aR509) is missing yet. So I added in this PR so this customData can be delivered from the outside of UIKit React.

### DIFF
--- a/scripts/index_d_ts
+++ b/scripts/index_d_ts
@@ -294,6 +294,7 @@ declare module "SendbirdUIKitGlobal" {
     isMessageReceiptStatusEnabledOnChannelList?: boolean;
     uikitOptions?: UIKitOptions;
     sdkInitParams?: SendbirdChatParams<Module[]>;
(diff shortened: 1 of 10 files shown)


sravan-s (approved):

AhyoungRyu:
I tested a bit with Storybook and confirmed it works after the last commit https://github.com/sendbird/sendbird-uikit-react/pull/698/commits/91c90ccef85759190ebaa555cc2afcae35a1ca70

PR #705, opened 2023-08-11
fix: remove duplicated getEmoji API call

AhyoungRyu:
Resolves https://sendbird.atlassian.net/browse/CLNP-542


src/modules/Channel/context/hooks/useGetChannel.js
```diff
@@ -39,7 +39,6 @@ function useSetChannel({ channelUrl, sdkInit, disableMarkAsRead }, {
       }).catch((err) => {
         logger.error('Channel: Getting emojis failed', err);
       });
-      sdk.getAllEmoji();
```
AhyoungRyu, line 42: The same API call is in the right above.

HoonBaek (approved):

PR #708, opened 2023-08-11
feat: Provide fetchChannelList method via ChannelListContext

HoonBaek:
### What was the issue
Customers could fetch the channel list using the `channelSource` of ChannelListContext
but it couldn't be able to update the `allChannels` because the state is only updated by an internal dispatcher.

### Fix
Add fetchChannelList to the ChannelListContext
* Make a custom hook function `useFetchChanneelList`
  * Use it for fetching channel list inside of the ChannelListUI component
  * Add tests for this function
* Provide the method through the ChannelListContext: fetchChannelList

### Usage
Here's an example of using `getFetchChannelList`

1. import below
```javascript
import SendbirdProvider from '@sendbird/uikit-react/SendbirdProvider'
import useSendbirdStateContext from '@sendbird/uikit-react/useSendbirdStateContext'
import { ChannelListProvider, useChannelListContext } from '@sendbird/uikit-react/ChannelList/context'
```
2. implement a custom channel list
```javascript
const isAboutSame = (a, b, px) => (Math.abs(a - b) <= px);
export const CustomChannelList = () => {
  const {
    allChannels,
    fetchChannelList,
  } = useChannelListContext();

  <div
    className="custom-channel-list"
    onScroll={(e) => {
      const target = e.target;
      if (isAboutSame(target.clientHeight + target.scrollTop, target.scrollHeight, 10)) {
        fetchChannelList();
      }
    }}
  >
    {allChannels.map((channel) => {
      return // custom channel list item
    })}
  </div>
}
```
3. apply it to the custom app
```javascript
import { CustomChannelList } from '../customChannelList.jsx'

const CustomApp = () => {
  return (
    <div className="custom-app">
      <SendbirdProvider ... >
        <ChannelListProvider ... >
          <CustomChannelList />
        </ChannelListProvider>
      </SendbirdProvider>
    </div>
  )
}
```

[CLNP-348](https://sendbird.atlassian.net/browse/CLNP-348)

[CLNP-348]: https://sendbird.atlassian.net/browse/CLNP-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/selectors.ts
```diff
@@ -653,6 +655,32 @@ export const getResendFileMessage = (state: SendBirdState) => (
...
+      logger,
+      markAsDeliveredScheduler,
+      disableMarkAsDelivered,
+    } = state.config;
+    const {
+      channelSource,
+      channelListDispatcher,
+    } = context;
+
+    useFetchChannelList({
```
AhyoungRyu, line 671: It looks a bit awkward to call hook from the normal js function since it's not neither React component nor custom hook. https://react.dev/warnings/invalid-hook-call-warning#breaking-rules-of-hooks

src/modules/ChannelList/context/hooks/useFetchChannelList.ts
```diff
@@ -0,0 +1,71 @@
...
+  channelSource: Nullable<GroupChannelListQuery>;
+  premiumFeatureList: string[];
+  disableMarkAsDelivered: boolean;
+}
+interface StaticProps {
+  channelListDispatcher: CustomUseReducerDispatcher;
+  logSubtitle: string;
+  logger: Logger;
+  markAsDeliveredScheduler: MarkAsDeliveredSchedulerType;
+}
```
liamcho, line 20: dont we need to export DynamicProps and StaticProps? also name is not specific

src/index.d.ts
```diff
@@ -273,6 +277,7 @@ interface sendbirdSelectorsInterface {
   getDeleteMessage: (store: SendBirdState) => GetDeleteMessage;
   getResendUserMessage: (store: SendBirdState) => GetResendUserMessage;
   getResendFileMessage: (store: SendBirdState) => GetResendFileMessage;
+  getFetchChannelList: (store: SendBirdState) => GetFetchChannelListType;
```
sravan-s, line 280: can you add usage example to PR description?

src/lib/selectors.ts
sravan-s: It will be called inside `React component `, but yeah, agree -> it customer would put this inside loop/conditional or something and potentially could break the rule of hook with this ~

sravan-s: @HoonBaek , probably better to export the hook through `sendbird/uikit-react/Channel/context`

src/modules/ChannelList/context/hooks/useFetchChannelList.ts
HoonBaek: am... the props are not re-used in the other places, so I think we don't need to export for now

src/lib/selectors.ts
sravan-s: Me and @HoonBaek  discussed, and we wont be exporting this as `sendbirdSelector`
We will add it to to the `ChannelListContext` and customer can use it through there


BTW ~ 
@HoonBaek When you have time, lets add documentation PR ~

AhyoungRyu: Yeah I just saw Hoon's message in Slack too. Like the idea 👍

HoonBaek: applied it~!

src/modules/ChannelList/context/hooks/useFetchChannelList.ts
```diff
@@ -0,0 +1,69 @@
...
+  channelSource,
+  premiumFeatureList,
+  disableMarkAsDelivered,
+}: DynamicProps, {
+  channelListDispatcher,
+  logger,
+  markAsDeliveredScheduler,
+}: StaticProps) => {
+  return useCallback(async () => {
+    const setMarkAsDelivered = premiumFeatureList.find((feature) => feature === DELIVERY_RECEIPT);
```
sravan-s, line 31: ideally this shoudlnt be calculated isndie this hook, but we can let it go for now 😉

sravan-s (approved):
Wait for one more review

AhyoungRyu (approved):

PR #712, opened 2023-08-17
Combine single and multi file viewer

sravan-s:
This is a n example to merge single/multi file viewer ~
There are two ways to do this, and this is one ->
Option 1. As in this PR, I made a `ViewerTypes` flag to deterministically detect the kind of props to recive(If single/undefined -> we assume single file, for backward compactability. Multi has to be explicitly flagged)
Option 2. We deduce ViewerTypes from list of files in the Message -> if there is only one, assume it is single, otherwise multi

Anyways, generic idea is -> calculate the current picture inside ``useViewerState``
Keep click handlers in separate hook..

### DIFF
--- a/src/ui/FileViewer/hooks/useViewerState.ts
+++ b/src/ui/FileViewer/hooks/useViewerState.ts
@@ -1,0 +1,54 @@
+import { useState } from "react";
+import { FileViewerComponentProps, ViewerTypes } from "../types";
+import { noop } from "../../../utils/utils";
+import { mapProps } from "./mapProps";
+
+export function useViewerState({ props }: { props: FileViewerComponentProps }): {
+  idx: number,
+  incrementIdx: () => void,
+  decrementIdx: () => void,
+  hasNext: boolean,
+  hasPrev: boolean,
+  name: string,
+  // maybe rename to mimeType
+  type: string,
+  url: string,
+} {
+  // you can keep the file info list in state
+  // todo - reset message idx on close
+  const [idx, setIdx] = useState(0);
+  const { name, type, url } = mapProps({ props, idx });
+  if (props.viewerType === ViewerTypes.MULTI) {
+    return {
+      idx,
+      incrementIdx: () => {
+        if (idx === props.fileInfoList.length - 1) {
+          return;
+        }
+        setIdx((prev) => prev + 1);
+      },
+      decrementIdx: () => {
+        if (idx === 0) {
+          return;
+        }
+        setIdx((prev) => prev - 1);
+      },
+      // todo - add more state
+      hasNext: false, // boolean -> implement
+      hasPrev: false, // boolean -> implements
+      name, // string
+      type, // string
+      url, // string
+    };
+  }
+  return {
+    idx,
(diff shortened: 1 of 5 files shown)


src/ui/FileViewer/types.ts
```diff
@@ -32,3 +35,41 @@ export const unSupported = (type: SupportedMimesType): boolean => (
 );
 
 export default { ...SUPPORTED_MIMES };
```
sravan-s, line 37: maybe move this to the end of the file?

src/ui/FileViewer/hooks/useViewerState.ts
```diff
@@ -0,0 +1,54 @@
...
+      type, // string
+      url, // string
+    };
+  }
+  return {
+    idx,
+    incrementIdx: noop,
+    decrementIdx: noop,
+    hasNext: false,
+    hasPrev: false,
```
sravan-s, line 49: These are default values for `SINGLE`

src/ui/FileViewer/types.ts
```diff
@@ -32,3 +35,41 @@ export const unSupported = (type: SupportedMimesType): boolean => (
...
+  // this is a type guard
+  viewerType?: typeof ViewerTypes.SINGLE,
+  isByMe?: boolean;
+  disableDelete?: boolean;
+  onClose: (e: MouseEvent) => void;
+  onDelete: (e: MouseEvent) => void;
+}
+
+export interface MultiFileViewer extends SenderInfo {
+  viewerType: typeof ViewerTypes.MULTI;
```
liamcho, line 64: does the caller needs to give this as input for MULTI?

src/ui/FileViewer/index.tsx
```diff
@@ -155,6 +170,21 @@ export default function FileViewer({
   onClose,
   onDelete,
 }: FileViewerProps): ReactElement {
+  // check if the message is single/multi
+  if (message?.isMultipleFilesMessage()) {
+    return (
+      <FileViewerComponent
+        profileUrl={message?.sender?.profileUrl}
+        nickname={message?.sender?.nickname}
+        viewerType={ViewerTypes.MULTI}
```
liamcho, line 179: So this part where customer has to give ViewerTypes.MULTI is what I don't like becuase customer would expect when giving fileInfoList would make this for MultiFilesViewer automatically.

sravan-s: yeah, you can rewrite that way too

src/ui/FileViewer/index.tsx
```diff
@@ -1,46 +1,54 @@
...
+    ref,
+    incrementIdx,
+    decrementIdx,
+    onClose: onClose || noop,
+  });
   const { stringSet } = useContext(LocalizationContext);
+
   return (
-    <div className="sendbird-fileviewer">
+    <div className="sendbird-fileviewer" ref={ref}>
```
liamcho, line 51: definitely a better solution 👍

liamcho (approved):
LGTM

PR #724, opened 2023-08-30
fix: Improve util functions type

HoonBaek:


src/utils/index.ts
```diff
@@ -415,18 +415,26 @@ export const hasSameMembers = <T>(a: T[], b: T[]): boolean => {
...
 
 export const filterMessageListParams = (
   params: MessageListParams,
   message: SendableMessageType,
 ): boolean => {
+  const {
+    customTypesFilter = [],
+  } = params as MessageListParams & {
+    customTypesFilter: Array<string>;
+  };
```
HoonBaek, line 431: Does this look okay?
@bang9

bang9: I believe we don't need `as MessageListParams & {
    customTypesFilter: Array<string>;
  };`

HoonBaek: Oh, that's true, thanks!

AhyoungRyu (approved):

src/utils/index.ts
```diff
@@ -320,7 +320,7 @@ export const isMessageSentByMe = (
   userId: string,
   message: SendableMessageType,
 ): boolean => (
-  (userId && message?.sender?.userId) && userId === message.sender.userId
+  !!(userId && message?.sender?.userId && userId === message.sender.userId)
```
liamcho, line 323: why do we need to do this?

PR #729, opened 2023-09-12
fix: use messageOffsetTimestamp instead of joinedAt

HoonBaek:
[CLNP-966](https://sendbird.atlassian.net/browse/CLNP-966)

* Remove replyType === 'THREAD' because the quoted message should show the unavailable message regardless of the reply type.
* Use messageOffsetTimestamp instead of joinedAt because joinedAt can't cover the every case

[CLNP-966]: https://sendbird.atlassian.net/browse/CLNP-966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/MessageContent/index.tsx
+++ b/src/ui/MessageContent/index.tsx
@@ -269,7 +269,7 @@ export default function MessageContent({
               message={message as SendableMessageType}
               userId={userId}
               isByMe={isByMe}
-              isUnavailable={(channel?.messageOffsetTimestamp ?? 0) * 1000 > (message.parentMessage?.createdAt ?? 0)}
+              isUnavailable={(channel?.messageOffsetTimestamp ?? 0) > (message.parentMessage?.createdAt ?? 0)}
               onClick={() => {
                 if (replyType === 'THREAD' && threadReplySelectType === ThreadReplySelectType.THREAD) {
                   onQuoteMessageClick?.({ message: message as SendableMessageType });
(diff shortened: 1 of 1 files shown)


src/ui/MessageContent/index.tsx
```diff
@@ -269,7 +269,7 @@ export default function MessageContent({
               message={message as SendableMessageType}
               userId={userId}
               isByMe={isByMe}
-              isUnavailable={(replyType === 'THREAD' && (channel?.joinedAt ?? 0 * 1000) > (message.parentMessage?.createdAt ?? 0))}
+              isUnavailable={(channel?.messageOffsetTimestamp ?? 0) * 1000 > (message.parentMessage?.createdAt ?? 0)}
```
HoonBaek, line 272: - Remove `replyType === 'THREAD'` because the quoted message should show the unavailable message regardless of the reply type.
- Use `messageOffsetTimestamp` instead of `joinedAt` because joinedAt can't cover the every case

bang9, line 272: It seems `* 1000` no needed
image

bang9: FYI
image

liamcho: yes you are right and we checked too! I think @HoonBaek forgot to push it

liamcho, line 272:
```suggestion
              isUnavailable={(channel?.messageOffsetTimestamp ?? 0) > (message.parentMessage?.createdAt ?? 0)}
```

HoonBaek: Ah, yes.
Liam and I found it after I pushed this PR, and I haven't updated it yet.
Thanks for reminding me!

liamcho (approved):
lg2m

PR #731, opened 2023-09-12
feat: Support MFM quote message input

liamcho:
Screen Shot 2023-09-12 at 5 27 52 PM

### DIFF
--- a/src/ui/QuoteMessage/utils.ts
+++ b/src/ui/QuoteMessage/utils.ts
@@ -5,10 +5,10 @@ import {Thumbnail} from "@sendbird/chat/lib/__definition";
 
 export function getMessageFirstFileType(message: CoreMessageType): string {
   return match(message)
-    .when(m => isFileMessage(m), () => {
+    .when(isFileMessage, () => {
       return (message as FileMessage)?.type ?? '';
     })
-    .when(m => isMultipleFilesMessage(m), () => {
+    .when(isMultipleFilesMessage, () => {
       const castedMessage: MultipleFilesMessage = message as MultipleFilesMessage;
       return getFirstFileInfo(castedMessage)?.mimeType ?? '';
     })
--- a/src/ui/QuoteMessage/utils.ts
+++ b/src/ui/QuoteMessage/utils.ts
@@ -17,7 +17,7 @@ export function getMessageFirstFileType(message: CoreMessageType): string {
     });
 }
 
-function getFirstFileInfo(message: MultipleFilesMessage): UploadedFileInfo {
+function getFirstFileInfo(message: MultipleFilesMessage): UploadedFileInfo | null {
   const fileInfoList: UploadedFileInfo[] = message.fileInfoList;
   return fileInfoList.length > 0 ? fileInfoList[0] : null;
 }
--- a/src/ui/QuoteMessage/utils.ts
+++ b/src/ui/QuoteMessage/utils.ts
@@ -9,8 +9,7 @@ export function getMessageFirstFileType(message: CoreMessageType): string {
       return (message as FileMessage)?.type ?? '';
     })
     .when(isMultipleFilesMessage, () => {
-      const castedMessage: MultipleFilesMessage = message as MultipleFilesMessage;
-      return getFirstFileInfo(castedMessage)?.mimeType ?? '';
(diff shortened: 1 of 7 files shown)


src/ui/QuoteMessage/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+import {match} from "ts-pattern";
+import {FileMessage, MultipleFilesMessage, UploadedFileInfo} from "@sendbird/chat/message";
+import {Thumbnail} from "@sendbird/chat/lib/__definition";
+
+export function getMessageFirstFileType(message: CoreMessageType): string {
+  return match(message)
+    .when(m => isFileMessage(m), () => {
+      return (message as FileMessage)?.type ?? '';
+    })
+    .when(m => isMultipleFilesMessage(m), () => {
```
AhyoungRyu, line 11: nit; this works same. 

```suggestion
    .when(isFileMessage, () => {
      return (message as FileMessage)?.type ?? '';
    })
    .when(isMultipleFilesMessage, () => {
```

src/ui/QuoteMessage/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+      const castedMessage: MultipleFilesMessage = message as MultipleFilesMessage;
+      return getFirstFileInfo(castedMessage)?.mimeType ?? '';
+    })
+    .otherwise(() => {
+      return '';
+    });
+}
+
+function getFirstFileInfo(message: MultipleFilesMessage): UploadedFileInfo {
```
AhyoungRyu, line 20:
```suggestion
function getFirstFileInfo(message: MultipleFilesMessage): UploadedFileInfo | null {
```
It seems it returns `null` too.

src/ui/QuoteMessage/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export function getMessageFirstFileName(message: CoreMessageType): string {
+  return match(message)
+    .when(m => isFileMessage(m), () => {
+      const castedMessage: FileMessage = message as FileMessage;
+      return castedMessage?.name ?? '';
+    })
+    .when(m => isMultipleFilesMessage(m), () => {
+      const castedMessage: MultipleFilesMessage = message as MultipleFilesMessage;
+      return getFirstFileInfo(castedMessage)?.fileName ?? '';
```
AhyoungRyu, line 33:
```suggestion
      return (message as FileMessage)?.name ?? '';
    })
    .when(isMultipleFilesMessage, () => {
      return (message as MultipleFilesMessage)?.fileName ?? '';
```

Like you did at the above in `getMessageFirstFileType`,  we can make it simpler in that way. 
Same for `getMessageFirstFileUrl` or `getMessageFirstFileThumbnails`.

src/utils/index.ts
```diff
@@ -239,6 +241,8 @@ export const isEnabledOGMessage = (message: UserMessage): boolean => (
 );
 
 export const getUIKitMessageTypes = (): UIKitMessageTypes => ({ ...UIKitMessageTypes });
+
+// Do not use this for MultipleFilesMessage. Use isMultipleFilesMessage() instead.
```
AhyoungRyu, line 245:
```suggestion
/**
 * Do not use this for MultipleFilesMessage. Use isMultipleFilesMessage() instead.
 */
```
nit;
To give a hint in IDE, the comment format should be like the above. 
Screenshot 2023-09-12 at 5 55 13 PM

src/ui/QuoteMessage/utils.ts
liamcho: oh nice, thanks!

src/ui/QuoteMessage/utils.ts
liamcho: thanks for the catch!

src/ui/QuoteMessage/utils.ts
liamcho: ah nice suggestion! castedMessage is no longer needed after replacing repeating code with `getFirstFileInfo`.

src/utils/index.ts
liamcho: nice!

HoonBaek (approved):
LGTM

ps. Let's migrate the functions `is~~~Message` and the function `getUIKitMessageType` ASAP

PR #733, opened 2023-09-13
feat: Create a GlobalModalProvider

HoonBaek:
## Fixes
* Remove the legacy of useModal, because it has not been used.
   : useModal, Modal component, ModalHeader comp, ...
* Make the other files forward the `useModal/index.tsx` instead of `useModal/ModalRoot`
   when they try to import MODAL_ROOT
* Add a props `onClose` to the component `ui/Modal`
  * The `onCancel` props should be deprecated in the next major version
* Create a `GlobalModalProvider` component

### How to use the GlobalModalProvider?
```tsx
import React, { useCallback } from 'react';
import { GlobalModalProvider, useGlobalModalContext } from 'src/hooks/useModal';

export const CustomComponent = () => {
  const { openModal } = useGlobalModalContext();

  const showUpAlert = useCallback(() => {
    openModal({
      modalProps: {},
      childElement: (
        <div className="custom-alert">
          "Button has been clicked!"
        </div>
      ),
    });
  }, []);

  return (
    <div className="custom-button">
      <button onClick={showUpAlert}>OPEN</button>
    </div>
  );
};

const CustomApp = () => {
  return (
    <div className="custom-app" styles={{ ... }}>
      <GlobalModalProvider>
        <CustomComponent />
      </GlobalModalProvider>
    </div>
  )
}
```


liamcho:
@HoonBaek 
The example should be
```
<button onClick={showUpAlert}>OPEN</button>
```
right?

scripts/index_d_ts
```diff
@@ -2427,16 +2427,20 @@ declare module '@sendbird/uikit-react/ui/MessageStatus' {
...
+    hideFooter?: boolean;
+    disabled?: boolean;
+    type?: string;
     isFullScreenOnMobile?: boolean;
     isCloseOnClickOutside?: boolean;
-    children: React.ReactElement | string,
-    renderHeader?: () => React.ReactElement,
+    children: React.ReactElement | string;
+    /**
+     * Do not use this! onCancel will be deprecated. Use onClose instead.
```
liamcho, line 2438: Should we include in which version it will be deprecated?

HoonBaek: Let's say v4, is it okay? @liamcho

src/hooks/useModal/README.md
```diff
@@ -1,24 +1,27 @@
...
+          "Button has been clicked!"
+        </div>
+      ),
+    });
+  }, []);
+
   return (
-    <div>
+    <div className="custom-button">
       <button onClick={open}>OPEN</button>
```
liamcho, line 23: where is `open`?

HoonBaek: This is my mistake :D I replace it with the `showUpAlert`

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+const GlobalModalContext = createContext<GlobalModalContextInterface>({
+  openModal: noop,
+});
+
+export const GlobalModalProvider = ({ children }: GlobalModalProviderProps) => {
+  // Idea from https://dev.to/nurislamov/simple-modals-queue-in-react-4g6c
+  const [globalModalQueue, setGlobalModalQueue] = useState<Array<OpenGlobalModalProps>>([]);
+
+  const openModal = useCallback((props: OpenGlobalModalProps) => {
+    setGlobalModalQueue((currentQue) => [...currentQue, props]);
```
liamcho, line 29: just curious, why is this better than
```
setGlobalModalQueue([... globalModalQueue, props]);
```
?

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  openModal: (props: OpenGlobalModalProps) => void;
+}
+
+const GlobalModalContext = createContext<GlobalModalContextInterface>({
+  openModal: noop,
+});
+
+export const GlobalModalProvider = ({ children }: GlobalModalProviderProps) => {
+  // Idea from https://dev.to/nurislamov/simple-modals-queue-in-react-4g6c
+  const [globalModalQueue, setGlobalModalQueue] = useState<Array<OpenGlobalModalProps>>([]);
```
liamcho, line 26: Which is better?
```
useState<OpenGlobalModalProps[]>
```

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const ModalComponent = useMemo(() => () => {
+    return match(globalModalQueue)
+      .when(q => q.length === 0, () => {
+        return <></>;
+      })
+      .otherwise(() => {
+        const { modalProps, childElement } = globalModalQueue[0];
+        return createPortal(
+          <Modal
+            {...{
```
liamcho, line 45: why we need `...` here?

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+              ...modalProps,
+              className: `sendbird-global-modal ${modalProps?.className}`,
+              onClose: () => {
+                modalProps?.onClose?.();
+                closeModal();
+              },
+            }}
+          >
```
liamcho, line 53: Can we do
```
          <Modal
            {...modalProps}
            className: `sendbird-global-modal ${modalProps?.className}`
            onClose: () => {
              modalProps?.onClose?.();
              closeModal();
            }
          >
```
?

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }, []);
+
+  const ModalComponent = useMemo(() => () => {
+    return match(globalModalQueue)
+      .when(q => q.length === 0, () => {
+        return <></>;
+      })
+      .otherwise(() => {
+        const { modalProps, childElement } = globalModalQueue[0];
+        return createPortal(
```
liamcho, line 43: what does this do?

src/hooks/useModal/index.tsx
HoonBaek: It's to make use of Functional Updates.

There are several advantages to this approach:
1. Safe Updates: By using functional updates, you avoid potential issues related to relying on the current state (count) directly for updates. This helps prevent unexpected behavior in cases of asynchronous updates or consecutive updates.
2. Optimization: React can optimize component rendering when you use functional updates. React keeps track of the previous state values between renders, and when you use functional updates, it can avoid unnecessary re-renders, thus improving performance.
3. Readability: Functional updates use functions, which can make complex state update logic more explicit and easier to trace and debug.

src/hooks/useModal/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Modal>,
+          document.getElementById(MODAL_ROOT) as HTMLElement,
+        );
+      });
+  }, [globalModalQueue]);
+
+  return (
+    <GlobalModalContext.Provider
+      value={{
+        openModal,
```
liamcho, line 64: not sure how this works but it seems useGlobalModalContext returns openModal?

src/hooks/useModal/index.tsx
HoonBaek: am... we decided to not use this pattern in this project, but it's long time ago
```
type[]
```

src/hooks/useModal/index.tsx
HoonBaek: Yes! this looks much better! I will apply it

src/hooks/useModal/index.tsx
HoonBaek: Oops, it shouldn't be here! let me remove it
It's already inside of the `<Modal />` component!

src/hooks/useModal/index.tsx
HoonBaek: Yes, it does.
Let's see if this really works in the next PR :)

src/hooks/useModal/index.tsx
liamcho: Nice explanation, thanks! Definitely good reasoning especially 1 and 2.

liamcho (approved):
Nice PR, LGTM!

HoonBaek:
Thank you for the good reviews @liamcho!

PR #734, opened 2023-09-13
fix: Error - properties of undefined in MessageSearch

HoonBaek:
fix: error -> Cannot read properties of undefined (reading ‘match’) in MessageSearch module


liamcho (approved):
LG2M

PR #738, opened 2023-09-15
bug-fix: Fixed a bug in MobileLayout where panel is not switched to channel list from channel when current user is banned or channel is deleted.

liamcho:
Fixed a bug in MobileLayout where panel is not switched to channel list from channel when current user is banned or channel is deleted.

Fixes: [UIKIT-3999](https://sendbird.atlassian.net/browse/UIKIT-3999)

[UIKIT-3999]: https://sendbird.atlassian.net/browse/UIKIT-3999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HoonBaek (approved):
👍

PR #740, opened 2023-09-18
bug-fix: Do not support MFM in super or broadcast channels.

liamcho:
Do not support MFM in super or broadcast channels.

Fixes: [CLNP-1045](https://sendbird.atlassian.net/browse/CLNP-1045)

[CLNP-1045]: https://sendbird.atlassian.net/browse/CLNP-1045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/MessageInput/utils.js
+++ b/src/ui/MessageInput/utils.js
@@ -38,4 +38,6 @@ export const nodeListToArray = (childNodes) => {
   }
 };
 
-export default debounce;
+export function isChannelTypeSupportsMultipleFilesMessage(channel) {
+  return channel.isGroupChannel() && !channel.isBroadcast && !channel.isSuper;
+}
--- a/src/ui/MessageInput/utils.js
+++ b/src/ui/MessageInput/utils.js
@@ -39,5 +39,9 @@ export const nodeListToArray = (childNodes) => {
 };
 
 export function isChannelTypeSupportsMultipleFilesMessage(channel) {
-  return channel.isGroupChannel() && !channel.isBroadcast && !channel.isSuper;
+  return channel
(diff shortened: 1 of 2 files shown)


src/ui/MessageInput/utils.js
```diff
@@ -38,4 +38,10 @@ export const nodeListToArray = (childNodes) => {
   }
 };
 
-export default debounce;
+export function isChannelTypeSupportsMultipleFilesMessage(channel) {
+  return channel
+    && typeof channel.isGroupChannel === 'function'
+    && channel.isGroupChannel()
```
AhyoungRyu, line 44: nit; 
I think those two lines can be simplified like below. 
```suggestion
    && channel?.isGroupChannel?.()
```

src/ui/MessageInput/utils.js
```diff
@@ -38,4 +38,10 @@ export const nodeListToArray = (childNodes) => {
   }
 };
 
-export default debounce;
```
AhyoungRyu, line 41: Doesn't it have any references?

src/ui/MessageInput/utils.js
liamcho: oh thanks! `&& channel?.isGroupChannel?()` I tried this and didnt work but now I understand `.` was missing.

src/ui/MessageInput/utils.js
liamcho: no reference so I removed it

AhyoungRyu (approved):
👍

PR #743, opened 2023-09-18
fix: apply appInfo.uploadSizeLimit to the files selection logic

HoonBaek:
* Add `uikitUploadSizeLimit` to the SendbirdProvider.config
* It is from appInfo.uploadSizeLimit
* The default value is 26 MB.


liamcho (approved):
LGTM!

PR #751, opened 2023-09-22
feat: Divide message list in the MessagesStore

HoonBaek:
### Issue
When succeding to sending multiple messages in a very short time, some of the pending messages are not removed from the message list.
We anticipate this is a timing issue.

### Fix
* To operate the message list safely, we are going to make another message list `localMessages` inside of the MessagesStore.
* From the next minor version `v3.7.0`, the `allMessages` will be deprecated or changed.
  * The current `allMessages` will only include the 'succeeded' messages, and the `localMessages` will contain the 'pending' and 'failed' messages.


liamcho (approved):
LG2M!

PR #755, opened 2023-09-25
fix: add supported mime types in the message input

HoonBaek:
* fix: add supported mime types in the message input
for the `text` and `application` mime types, and also add `.mov` to the video


liamcho (approved):
Nice job~ LGTM

PR #763, opened 2023-10-04
fix: wrong isResendable cond check in openChannel/useResendMessageCallback 

AhyoungRyu:
~Fixes https://sendbird.atlassian.net/browse/CLNP-905~

`message.isResendable` type is function, and its return type is boolean.  So the way we've checked the value like 
```
if (failedMessage?.isResendable) { ... }
```

has been wrong but it should be if we really want to check its existence. 
```
if (typeof failedMessage?.isResendable === 'boolean') { ... }
```

### DIFF
--- a/src/modules/Channel/context/hooks/useResendMessageCallback.js
+++ b/src/modules/Channel/context/hooks/useResendMessageCallback.js
@@ -11,7 +11,7 @@ function useResendMessageCallback({
   return useCallback((failedMessage) => {
     logger.info('Channel: Resending message has started', failedMessage);
     const { messageType, file } = failedMessage;
-    if (failedMessage?.isResendable) {
+    if (typeof failedMessage?.isResendable === 'boolean') {
       // Move the logic setting sendingStatus to pending into the reducer
       // eslint-disable-next-line no-param-reassign
       failedMessage.requestState = 'pending';
--- a/src/modules/Channel/context/hooks/useResendMessageCallback.js
+++ b/src/modules/Channel/context/hooks/useResendMessageCallback.js
@@ -11,7 +11,7 @@ function useResendMessageCallback({
   return useCallback((failedMessage) => {
     logger.info('Channel: Resending message has started', failedMessage);
     const { messageType, file } = failedMessage;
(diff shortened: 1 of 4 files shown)


AhyoungRyu:
Wait, I need to check a bit more.

src/modules/OpenChannel/context/hooks/useResendMessageCallback.ts
```diff
@@ -25,7 +25,7 @@ function useResendMessageCallback(
     logger.info('OpenChannel | useResendMessageCallback: Resending message has started', failedMessage);
     // eslint-disable-next-line no-param-reassign
     const { messageType, file } = failedMessage as FileMessage;
-    if (failedMessage && typeof failedMessage.isResendable === 'function' && failedMessage.isResendable) {
+    if (typeof failedMessage?.isResendable === 'boolean') {
```
liamcho, line 28:
```suggestion
    if (typeof failedMessage?.isResendable === 'boolean' && failedMessage.isResendable) {
```
?

src/modules/Channel/context/hooks/useResendMessageCallback.js
```diff
@@ -11,7 +11,7 @@ function useResendMessageCallback({
   return useCallback((failedMessage) => {
     logger.info('Channel: Resending message has started', failedMessage);
     const { messageType, file } = failedMessage;
-    if (failedMessage?.isResendable) {
+    if (typeof failedMessage?.isResendable === 'boolean') {
```
liamcho, line 14:
```suggestion
    if (typeof failedMessage?.isResendable === 'boolean' && failedMessage.isResendable) {
```
?

AhyoungRyu: Ah yeah, actually the original logic has no problem. Probably I don't even need the fix in this file at least. So I'm checking with Uday again.

AhyoungRyu:
The whole change in this file will not fix the linked SBIssue cause the SBIssue is related to group channel but at least, I fixed the incorrect logic in the open channel msg resend hook.

HoonBaek (approved):

PR #766, opened 2023-10-06
[CLNP-1150] fix: Textbox placeholder disappears when message is cleared in Safari

AhyoungRyu:
Resolves https://sendbird.atlassian.net/browse/SBISSUE-13846

This issue happens only in Safari.

#### AS-IS 
asis

If some texts are removed from the input box, the placeholder msg just disappears. 

#### TO-BE
tobe

With this patch, the placeholder msg should be returned if nothing left in the text field. This is the commonly expected behavior.

### DIFF
--- a/src/ui/MessageInput/index.jsx
+++ b/src/ui/MessageInput/index.jsx
@@ -397,7 +397,6 @@ const MessageInput = React.forwardRef((props, ref) => {
     setHeight,
   });
 
-  const textField = ref?.current;
   return (
     <form
       className={getClassName([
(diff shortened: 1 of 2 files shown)


src/ui/MessageInput/index.jsx
```diff
@@ -397,7 +397,6 @@ const MessageInput = React.forwardRef((props, ref) => {
     setHeight,
   });
 
-  const textField = ref?.current;
```
AhyoungRyu, line 400: Realized that saving the ref value in this way, it makes unexpected behavior e.g. the text value is not properly updated in realtime. So I removed this to make it the outdated value can't re-used.

HoonBaek (approved):

PR #777, opened 2023-10-16
fix: Improve the ChannelList reducer logic

HoonBaek:
[CLNP-1164](https://sendbird.atlassian.net/browse/CLNP-1164)

## Issue
It was not able to replace the current channel with another channel,
  when the current channel was filtered from the ChannelList AND there's no more channel to replace.

But we have tried to replace it with the other channels which don't exist.

### So I had to consider various and specific conditions for each case.
* onChannelForzen
* onChannelUnFrozen
  * Channel can be **added/removed/kept** by this event
  * The current channel should be replaced if it's removed
* onChannelChanged
  * Channel can be **added/removed/kept** by this event
  * The current channel should be replaced if it's removed
  * Channel can be moved to the top of the ChannelList
* onUserJoined
  * Channel can be **added or not** by this event
* onReadReceiptUpdated
  * Channel can be **added or not** by this event
* onUserLeft
  * Channel can be **removed or kept** by this event
  * The current channel should be replaced if it's removed
* onChannelHidden
  * Channel can be **added/removed/kept** by this event
  * The current channel should be replaced if it's removed
* creating a channel
  * Channel can be **added or not** by this event

### Mainly four conditions are considered
image

## Fix
* fix: replace the **currentChannel** if it's removed from the `ChannelList`
  * it can be filtered by the `channelListQuery` or because of the other events
    * e.g. **channel is hidden, the current user left the channel**
  * replace the **currentChannel** with the other channel
    * if `disableAutoSelect` is false, and there's more channel to replace in the `allChannels`
* fix: replace the old channel event handler in the `ChannelLIst`
  * **onReadStatus** to `onUnreadMemberStatusUpdated`
  * **onDeliveryReceiptUpdated** to `onUndeliveredMemberStatusUpdated`

[CLNP-1164]: https://sendbird.atlassian.net/browse/CLNP-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/ChannelList/dux/reducers.js
```diff
@@ -114,79 +124,98 @@ export default function reducer(state, action) {
...
+        const filteredChannel = channel;
+        let nextChannel = null;
+        if (currentChannel?.url === filteredChannel.url) {
+          if (!disableAutoSelect && allChannels.length > 0) {
+            const [firstChannel, secondChannel = null] = allChannels;
+            nextChannel = firstChannel.url === filteredChannel.url ? secondChannel : firstChannel;
+          }
+        } else {
+          nextChannel = currentChannel;
+        }
```
liamcho, line 193: I see this logic is used in three different places. Can we wrap it into a funciton?

HoonBaek: good suggestion, let me check the feasibility

src/modules/ChannelList/dux/reducers.js
```diff
@@ -42,63 +42,73 @@ export default function reducer(state, action) {
...
+
+      // No channelListQuery
+      // * Remove the channel from the ChannelList: because the channel is hidden
+      // Replace the currentChannel if it's filtered or hidden
+      let nextChannel = null;
+      if (currentChannel?.url === channel.url) {
+        if (!disableAutoSelect && allChannels.length > 0) {
+          const [firstChannel, secondChannel = null] = allChannels;
+          nextChannel = firstChannel.url === channel.url ? secondChannel : firstChannel;
+        }
```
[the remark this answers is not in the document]
HoonBaek: Oops, I think I misunderstood your comment, yes it's intended.

src/modules/ChannelList/dux/actionTypes.js
bang9, line 1: could you convert this to typescript? 🌝

src/modules/ChannelList/dux/reducers.js
```diff
@@ -42,63 +43,70 @@ export default function reducer(state, action) {
...
       const channel = action.payload;
-      if (state.channelListQuery) {
-        if (filterChannelListParams(state.channelListQuery, channel, state.currentUserId)) {
+      const {
+        allChannels,
+        currentUserId,
+        channelListQuery,
+      } = state;
+      if (channelListQuery) {
+        if (filterChannelListParams(channelListQuery, channel, currentUserId)) {
```
bang9, line 54: Channel event handlers are not affected by the filter, so should we manually filter this with query parameters, right?

src/modules/ChannelList/dux/getNextChannel.ts
```diff
@@ -0,0 +1,33 @@
+// See the conditions of this function here https://github.com/sendbird/sendbird-uikit-react/pull/777
+import { GroupChannel } from '@sendbird/chat/groupChannel';
+
+type CurrentChannelType = GroupChannel | null;
+
+export interface GetNextChannelProps {
```
bang9, line 6: how about params instead of props?

src/modules/ChannelList/dux/reducers.js
HoonBaek: Yes it is

src/modules/ChannelList/dux/getNextChannel.ts
```diff
@@ -0,0 +1,33 @@
...
+ * NOTICE: Use this function IF the current channel is removed from allChannels.
+ * This function will give you the next currentChannel value.
+ */
+export const getNextChannel = ({
+  channel,
+  currentChannel,
+  allChannels,
+  disableAutoSelect,
+}: GetNextChannelParams): CurrentChannelType => {
+  let nextChannel: CurrentChannelType = null;
```
AhyoungRyu, line 23: Do you mean? 
```suggestion
  let nextChannel: CurrentChannel = null;
```

src/modules/ChannelList/dux/reducers.js
```diff
@@ -114,79 +122,94 @@ export default function reducer(state, action) {
...
-          currentChannel: state.disableAutoSelect ? null : nextChannel,
-          allChannels: state.allChannels.filter(({ url }) => url !== channel?.url),
-        };
       }
-      const filteredChannels = state.allChannels.filter((c) => !(c.url === channel?.url && isMe));
-      const nextChannel = (isMe && (channel?.url === state.currentChannel?.url))
-        ? filteredChannels[0]
-        : state.currentChannel;
+      // Replace the currentChannel if I left the currentChannel
+      if (isMe && channel.url === currentChannel?.url) {
```
AhyoungRyu, line 141: It seems the similar logic is in the `getNextChannel` util function

src/modules/ChannelList/dux/getNextChannel.ts
```diff
@@ -0,0 +1,33 @@
+// See the conditions of this function here https://github.com/sendbird/sendbird-uikit-react/pull/777
+import { GroupChannel } from '@sendbird/chat/groupChannel';
+
+type CurrentChannelType = GroupChannel | null;
```
HoonBaek, line 4: @AhyoungRyu I define CurrentChannelType here

src/modules/ChannelList/dux/reducers.js
HoonBaek: Yes there is, but `isMe` is added, so little bit different

src/modules/ChannelList/dux/getNextChannel.ts
liamcho: CurrentChannel is not defined but
CurrentChannelType is defined above

src/modules/ChannelList/utils.js
```diff
@@ -34,18 +34,6 @@ const createEventHandler = ({
         });
       }
     },
-    onUserReceivedInvitation: (channel, inviter, invitees) => {
```
liamcho, line 37: why this is removed?

HoonBaek: Because it's not used, but I will add it again. There was a customer request.

src/modules/ChannelList/dux/reducers.js
```diff
@@ -42,63 +43,70 @@ export default function reducer(state, action) {
...
             ...state,
-            allChannels: getChannelsWithUpsertedChannel(state.allChannels, channel),
+            currentChannel: channel,
+            allChannels: getChannelsWithUpsertedChannel(allChannels, channel),
           };
         }
+        // Do not add to the ChannelList
         return {
           ...state,
           currentChannel: channel,
```
liamcho, line 65: no need to add allChannels here?

src/modules/ChannelList/dux/reducers.js
```diff
@@ -114,79 +122,94 @@ export default function reducer(state, action) {
...
+        currentChannel,
+        channelListQuery,
+        disableAutoSelect,
+      } = state;
+      let nextChannels = allChannels.filter((ch) => ch.url !== channel.url);
+      let nextChannel = null;
+      if (channelListQuery) {
+        if (filterChannelListParams(channelListQuery, channel, currentUserId)) {
+          // Good to [add to/keep in] the ChannelList
+          nextChannels = getChannelsWithUpsertedChannel(allChannels, channel);
```
liamcho, line 137: why upsert removed channel again here?

src/modules/ChannelList/dux/reducers.js
liamcho: She is right. It is redundant code. Let's replace with
```
let nextChannel;
if (isMe) {
  nextChannel = getNextChannel();
}
```

liamcho (commented):
Great job 👍  just check my comments

src/modules/ChannelList/dux/reducers.js
HoonBaek: Yes, because the channel is not appropriate, it's filtered

src/modules/ChannelList/dux/reducers.js
HoonBaek: It's not removed,
```
let nextChannels = allChannels.filter((ch) => ch.url !== channel.url);
```
This is just default value

liamcho (approved):
LG2M

PR #789, opened 2023-10-17
fix: remove unused files and fix ts settings

bang9:
- I removed unused settings and files, and renamed the used ones with clearer names.
- I modified the tsconfig.json file to properly recognize source files.
As a result, type errors that were supposed to occur in the source files will now appear.
This PR includes only some fixes for type errors. (we should fix this gradually) <br/>image


liamcho:
can you run locally `yarn lint:fix`?

AhyoungRyu (approved):
Thanks Airen for doing this!

PR #780, opened 2023-10-16
bug-fix: Allow delete failed messages in offline state (re-fix)

liamcho:
Fixes: [UIKIT-4520](https://sendbird.atlassian.net/browse/UIKIT-4520)

[UIKIT-4520]: https://sendbird.atlassian.net/browse/UIKIT-4520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #790, opened 2023-10-17
bug-fix: Allow delete failed messages in offline state (second re-fix)

liamcho:
Fixes: [UIKIT-4535](https://sendbird.atlassian.net/browse/UIKIT-4535)

[UIKIT-4535]: https://sendbird.atlassian.net/browse/UIKIT-4535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/MessageItemMenu/index.tsx
+++ b/src/ui/MessageItemMenu/index.tsx
@@ -13,11 +13,12 @@ import {
   isUserMessage,
   isSentMessage,
   isFailedMessage,
-  isPendingMessage, SendableMessageType, isMultipleFilesMessage,
+  isPendingMessage, SendableMessageType,
 } from '../../utils/index';
 import { LocalizationContext } from '../../lib/LocalizationContext';
 import { Role } from '../../lib/types';
 import { ReplyType } from '../../types';
+import { useChannelContext } from '../../modules/Channel/context/ChannelProvider';
 
 interface Props {
   className?: string | Array<string>;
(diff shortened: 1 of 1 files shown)


bang9:
Isn't it resolved by https://github.com/sendbird/sendbird-uikit-react/pull/780 ?

src/ui/MessageItemMenu/index.tsx
```diff
@@ -209,9 +211,13 @@ export default function MessageItemMenu({
...
-                    if (isMultipleFilesMessage(message) || !disabled) {
-                      showRemove(true);
-                      closeDropdown();
+                    if (!disabled || isFailedMessage(message)) {
+                      if (isFailedMessage(message)) {
+                        channelStore?.deleteMessage?.(message);
+                      } else {
+                        showRemove(true);
+                        closeDropdown();
+                      }
```
AhyoungRyu, line 220: Instead of using nested if stmt and repeat the same condition `isFiledMessage`, we can just do the early return like below 

```suggestion
                    if (disabled) {
                      return;
                    }
                    if (isFailedMessage(message)) {
                      channelStore?.deleteMessage?.(message);
                    } else {
                      showRemove(true);
                      closeDropdown();
                    }
```

liamcho: nice!

liamcho:
> Isn't it resolved by #780 ?

Nope. That PR was not enough to fix the issue. This PR is also needed.

bang9 (approved):

PR #795, opened 2023-10-18
feat: Added SBUEventHandlers and onPressUserProfile.

liamcho:
Fixes: [CLNP-1148](https://sendbird.atlassian.net/browse/CLNP-1148)

[CLNP-1148]: https://sendbird.atlassian.net/browse/CLNP-1148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/index.d.ts
+++ b/scripts/index.d.ts
@@ -118,6 +118,15 @@ declare module "SendbirdUIKitGlobal" {
     isMessageReceiptStatusEnabledOnChannelList?: boolean;
     uikitOptions?: UIKitOptions;
     sdkInitParams?: SendbirdChatParams<Module[]>;
+
+    // Customer provided callbacks
+    eventHandlers?: SBUEventHandlers;
+  }
+
+  export interface SBUEventHandlers {
+    reaction?: {
+      onPressUserProfile?(member: User): void;
+    },
   }
 
   export type Logger = {
@@ -301,6 +310,9 @@ declare module "SendbirdUIKitGlobal" {
     uikitOptions?: UIKitOptions;
     sdkInitParams?: SendbirdChatParams<Module[]>;
     customExtensionParams?: Record<string, string>;
+
+    // Customer provided callbacks
+    eventHandlers?: SBUEventHandlers;
   }
 
   export interface SendBirdStateConfig {
--- a/src/ui/MessageContent/index.tsx
+++ b/src/ui/MessageContent/index.tsx
@@ -108,7 +108,8 @@ export default function MessageContent({
 }: Props): ReactElement {
   const messageTypes = getUIKitMessageTypes();
   const { dateLocale } = useLocalization();
-  const { config } = useSendbirdStateContext?.() || {};
+  const { config, eventHandlers } = useSendbirdStateContext?.() || {};
+  const onPressUserProfileHandler = eventHandlers?.reaction?.onPressUserProfile;
   const { disableUserProfile, renderUserProfile } = useContext(UserProfileContext);
   const avatarRef = useRef(null);
   const contentRef = useRef(null);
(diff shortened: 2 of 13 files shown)


scripts/index.d.ts
```diff
@@ -118,6 +118,15 @@ declare module "SendbirdUIKitGlobal" {
...
     sdkInitParams?: SendbirdChatParams<Module[]>;
+
+    // Customer provided callbacks
+    eventHandlers?: SBUEventHandlers;
+  }
+
+  export interface SBUEventHandlers {
+    reaction?: {
+      onPressUserProfile?(member: User): void;
+    },
```
bang9, line 129: looks so fancy 👍

src/ui/MobileMenu/ReactedMembersBottomSheet.tsx
```diff
@@ -28,8 +29,17 @@ export const ReactedMembersBottomSheet = ({
   emojiContainer,
 }: ReactedMembersBottomSheetProps): ReactElement => {
   const { members = [] } = channel;
+  const globalState = useSendbirdStateContext?.();
```
bang9, line 32: @HoonBaek Is it safe to use `useSendbirdStateContext` in the `src/ui`?
I thought UI was an independent component element unrelated to business logic.

liamcho: Yes you are right. He was worried about this too. However, you can see that if `useSendbirdStateContext` is undefined, nothing happens (no error) so I think it is safe. If client wants this callback to work, they have to use `SendbirdProvider` too.

bang9: `useSendbirdStateContext` is an imported function and therefore always exists. Although the return value of `useSendbirdStateContext()` may not exist, changing the behavior of `useSendbirdStateContext` can lead to side effects.

For example, in a recent update, I modified it to throw an error if there's no Provider.
[Link to the modification](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/hooks/useSendbirdStateContext.tsx#L18)

If that's the policy(no business logic in ui), it seems like a good practice to continue following it for the sake of product maintenance and quality.

Would it be better to open the onPressUserProfileHandler interface in the UI?

```jsx
const EmojiReactions = ({ onPressUserProfile }) {
 return (<ReactedMembersBottomSheet onPressUserProfileHandler={onPressUserProfile} />)
}

const SomeModule = () => {
  const { eventHandlers } = useSendbirdStateContext();
  return (<EmojiReactions onPressUserProfileHandler={eventHandlers?.reaction?.onPressUserProfile} />)
}
```

liamcho: @bang9 that is good point. Let's not take this risk. Ill replace with onPressUserProfile

src/ui/MessageContent/index.tsx
```diff
@@ -108,7 +108,8 @@ export default function MessageContent({
 }: Props): ReactElement {
   const messageTypes = getUIKitMessageTypes();
   const { dateLocale } = useLocalization();
-  const { config } = useSendbirdStateContext?.() || {};
+  const { config, eventHandlers } = useSendbirdStateContext?.() || {};
```
bang9, line 111: This part is also a UI component, but since we're already using **useSendbirdStateContext**, let's skip it for now and make the changes later during refactoring ✨

bang9:
Questions

In relation to ticket, there are additional tasks besides the event handler:

1. Clicking on the profile within the reaction will change the default behavior to display the user profile. (Default: true)
2. The interface that displays the user profile should be made available to customers as a public interface.

Will these tasks be handled through a separate PR?

src/ui/MessageContent/index.tsx
liamcho: oh true. Let's keep this in mind.

liamcho:
@bang9 thanks for noticing this. I was not aware of 1 and 2. It will be in separate PRs and targeting their release next week's release but not today.

bang9 (approved):
lgtm

PR #801, opened 2023-10-19
chore: Fix missing pubsub PublishingModuleType

liamcho:


src/modules/internalInterfaces.ts
```diff
@@ -0,0 +1,4 @@
+export enum PublishingModuleType {
+  CHANNEL = 'CHANNEL',
+  THREAD = 'THREAD',
+}
```
HoonBaek, line 4: 👍

HoonBaek (approved):

PR #808, opened 2023-10-23
feat: apply localThreadMessages to the ThreadContext

HoonBaek:
Fix
* Divide the `allThreadMessages` into the `allThreadMessages` and `localThreadMessages`
  * The `localThreadMessages` includes the pending and failed messages
  * The `allThreadMessages` won't include those sending status messages


liamcho (approved):
LGTM

PR #810, opened 2023-10-23
chore: Fix scroll to bottom on reconnect.

liamcho:
Reelia-recording-2023-10-23


HoonBaek:
Can you explain the relationship between `onScrollDownCallback` and `reconnection`? @liamcho

HoonBaek (approved):
onScrollDownCallback is called when reconnecting

PR #811, opened 2023-10-23
fix: use useThreadContext instead useChannelContext in Thread

HoonBaek:
### Issue
Error when opening Thread component
> Cannot read properties of undefined (reading 'deleteMessage')

### Fix
* fix: use useThreadContext instead useChannelContext in Thread


liamcho (approved):
LGTM

PR #814, opened 2023-10-23
fix: pubsub subscriber in useEffect

HoonBaek:
### issue
* pubSub subscriber has been undefined, so couldn't remove the previous subscriptions
* delete command has been used to Array, but it's for Object

### Fix
* move the pubSub Subscriber into the useEffect sope for making effective
* use splice instead of delete in the pubSubFactory


liamcho (approved):
Nice fix! LGTM

PR #820, opened 2023-10-24
feat: new build settings

bang9:
Changelogs:

- Changes have been made to export modules using the [sub-path exports](https://nodejs.org/api/packages.html#subpath-exports) in the package.json.
If you were using the package in a Native CJS environment, this might have an impact.
You can migrate the path as follows:
  ```diff
  - const ChannelList = require('@sendbird/uikit-react/cjs/ChannelList');
  + const ChannelList = require('@sendbird/uikit-react/ChannelList');
  ```
 - TypeScript support has been improved.
Now, precise types based on the source code are used.

---

- Fix type errors in source.
- Exports auto generated .d.ts files instead of manual `index.d.ts` file

### Before
1. run rollup
    1. read inputs (module entries)
    2. generate css and js files to
        - css: `/dist/dist/index.css`, `/dist/dist/cjs/dist/index.css`
        - js: `/dist` and `/dist/cjs`
    3. copy `README.md`, `LICESNSE`, `CHANGELOG.md`, `index.d.ts`
 2. run `scripts/post_build.js`
     1. append version to `index.d.ts`
     2. package.json: add dependencies and copy to `dist/package.json`

### After
1. run rollup
    1. read inputs (module entries)
    2. generate css and js files to
        - css: `/dist/dist/index.css`, `/dist/dist/cjs/dist/index.css`
        - js: `/dist` and `/dist/cjs`
    3. copy `README.md`, `LICESNSE`, `CHANGELOG.md`
 2. run `scripts/post_build.js`
     1. package.json: add dependencies, exports, typesVersions and copy to `dist/package.json`
     2. remove unused css files in `/dist/dist/cjs/dist`
     3. build `.d.ts` files

---

The biggest difference is that now we use the exports supported in package.json.
You can find more information here: https://nodejs.org/api/packages.html#subpath-exports

This may result in breaking changes because you can only use modules that are specified in exports.
For example, the `@sendbird/chat/cjs/index.js` file actually exists, but it is not exported in package.json, so you cannot use it in a strict environment.

The advantages of this approach are as follows:
For paths that we directly specify in exports, files are assigned accordingly in the esm/cjs environment.
Previously, for individually exported entries (modules) other than the main entry, we had to specify the path directly depending on the module environment.

For example:
> #### Before
>esm: @sendbird/chat (package.json > module: "index.js")
cjs: @sendbird/chat (package.json > main: "cjs/index.js")
>
>esm: @sendbird/chat/ChannelList (nothing in package.json)
cjs: @sendbird/chat/cjs/ChannelList (nothing in package.json)
>
> #### After
>esm: @sendbird/chat (package.json > exports: { ".": "index.js" } )
cjs: @sendbird/chat (package.json > exports: { ".": "cjs/index.js" } )
>
>esm: @sendbird/chat/ChannelList (package.json > exports: { "./ChannelList": "ChannelList/index.js" } )
cjs: @sendbird/chat/ChannelList (package.json > exports: { "./ChannelList": "cjs/ChannelList/index.js" } )

---

package.json example
```json
{
  "exports": {
      "./package.json": "./package.json",
      "./dist/index.css": "./dist/index.css",
      ".": {
        "types": "./types/index.d.ts",
        "require": "./cjs/index.js",
        "import": "./index.js",
        "default": "./index.js"
      },
      "./App": {
        "types": "./types/modules/App/index.d.ts",
        "require": "./cjs/App.js",
        "import": "./App.js",
        "default": "./App.js"
      },
      "./SendbirdProvider": {
        "types": "./types/lib/Sendbird.d.ts",
        "require": "./cjs/SendbirdProvider.js",
        "import": "./SendbirdProvider.js",
        "default": "./SendbirdProvider.js"
      },
      "./sendbirdSelectors": {
        "types": "./types/lib/selectors.d.ts",
        "require": "./cjs/sendbirdSelectors.js",
        "import": "./sendbirdSelectors.js",
        "default": "./sendbirdSelectors.js"
      },
      "./VoiceRecorder/context": {
        "types": "./types/hooks/VoiceRecorder/index.d.ts",
        "require": "./cjs/VoiceRecorder/context.js",
        "import": "./VoiceRecorder/context.js",
        "default": "./VoiceRecorder/context.js"
      },
      "./VoiceRecorder/useVoiceRecorder": {
        "types": "./types/hooks/VoiceRecorder/useVoiceRecorder.d.ts",
        "require": "./cjs/VoiceRecorder/useVoiceRecorder.js",
        "import": "./VoiceRecorder/useVoiceRecorder.js",
        "default": "./VoiceRecorder/useVoiceRecorder.js"
      }
  }
}
```

```sh
git rebase --onto main HEAD~13
```


liamcho:
I have two Qs:

1.
> css: /dist/dist/index.css, /dist/dist/cjs/dist/index.css
Why can't it be 
css: /dist/index.css, /dist/cjs/dist/index.css
?

2. What is logic of `remove unused css files in /dist/dist/cjs/dist`?

babel.config.js
```diff
@@ -1,8 +1,21 @@
 module.exports = {
   presets: [
-    '@babel/preset-env',
+    [
+      '@babel/preset-env',
+      {
+        loose: true,
```
liamcho, line 6: what is this for?

package.json
```diff
@@ -132,19 +132,18 @@
...
-    "rollup-plugin-polyfill-node": "^0.10.1",
-    "rollup-plugin-postcss": "^4.0.0",
-    "rollup-plugin-scss": "^3.0.0",
-    "rollup-plugin-visualizer": "^5.9.0",
+    "rollup-plugin-copy": "^3.5.0",
+    "rollup-plugin-polyfill-node": "^0.12.0",
+    "rollup-plugin-postcss": "^4.0.2",
+    "rollup-plugin-scss": "^4.0.0",
+    "rollup-plugin-typescript2": "^0.36.0",
+    "rollup-plugin-visualizer": "^5.9.2",
```
liamcho, line 142: Nice but I hope nothing breaks after all these change.

scripts/post_build.js
```diff
@@ -3,70 +3,81 @@
...
-  */
-  const cleanedPkgJSON = {
-    version: json.version,
+async function createPackageJSON(json) {
+  const exportList = (await moduleExports).default;
+
+  const typesVersions = { '*': {} };
+  const exports = {
+    './package.json': './package.json',
+    './dist/index.css': './dist/index.css',
```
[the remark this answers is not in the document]
bang9: Since it's already being distributed that way, we need to maintain it (for backward compatibility).
If we update to v4, then we can move it to the top level at that time

scripts/post_build.js
```diff
@@ -3,70 +3,81 @@
...
-      typings: '../index.d.ts',
-    }, null, 2),
-    { flag: 'w' },
+    JSON.stringify({ type: 'commonjs' }, null, 2),
+    { flag: 'w' }
   );
 }
 
+function removeUnusedCSS() {
+  execSync('rm -rf dist/cjs/dist', { stdio: 'inherit' });
```
liamcho, line 70: This deletes all files. How do we know all files here are unused css files?

src/lib/dux/sdk/actionTypes.ts
```diff
@@ -10,7 +10,7 @@ export const SDK_ACTIONS = {
 
 type SDK_PAYLOAD_TYPES = {
   [SDK_ACTIONS.SET_SDK_LOADING]: boolean,
-  [SDK_ACTIONS.INIT_SDK]: SendbirdChat,
+  [SDK_ACTIONS.INIT_SDK]: SdkStoreStateType['sdk'],
```
liamcho, line 13: Nice

src/lib/selectors.ts
```diff
@@ -57,7 +57,7 @@ import { SendableMessageType } from '../utils';
 /**
  * const sdk = selectors.getSdk(state);
  */
-export const getSdk = (state: SendBirdState): SendbirdGroupChat & SendbirdOpenChat => {
+export const getSdk = (state: SendBirdState) => {
```
liamcho, line 60: Why remove return type?

src/lib/types.ts
```diff
@@ -107,8 +105,9 @@ export interface SdkStore {
   error: boolean;
   initialized: boolean;
   loading: boolean;
-  sdk: SendbirdChat & SendbirdGroupChat & SendbirdOpenChat;
+  sdk: SendbirdChat & ModuleNamespaces<[GroupChannelModule, OpenChannelModule]>;
```
liamcho, line 108: nice

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -173,7 +173,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
               type={LabelTypography.SUBTITLE_2}
               color={LabelColors.ONBACKGROUND_2}
             >
-              {stringSet.MENTION_COUNT__OVER_LIMIT.replace('%d', maxMentionCount)}
+              {stringSet.MENTION_COUNT__OVER_LIMIT.replace('%d', String(maxMentionCount))}
```
liamcho, line 176: nice

src/modules/Channel/components/TypingIndicator.tsx
```diff
@@ -15,27 +14,31 @@ export interface TypingIndicatorTextProps {
   members: Member[];
 }
 
-export const TypingIndicatorText: React.FC<TypingIndicatorTextProps> = ({ members }: TypingIndicatorTextProps) => {
-  const { stringSet } = useContext(LocalizationContext);
-  if (!members || members.length === 0) {
-    return '';
-  }
+export const TypingIndicatorText = ({ members }: TypingIndicatorTextProps) => {
```
liamcho, line 17: can we keep return type?

src/modules/CreateChannel/components/InviteUsers/index.tsx
```diff
@@ -21,10 +21,11 @@ import {
   createDefaultUserListQuery,
 } from './utils';
 import { noop } from '../../../../utils/utils';
+import { UserListQuery } from '../../../../types';
 
 export interface InviteUsersProps {
   onCancel?: () => void;
-  userListQuery?(): ApplicationUserListQuery;
+  userListQuery?(): UserListQuery;
```
liamcho, line 28: This is for inviting all application users to the new channel so it should be ApplicationUserListQuery no?

src/modules/MessageSearch/context/hooks/useSetChannel.ts
```diff
@@ -1,16 +1,16 @@
...
 interface MainProps {
   channelUrl: string;
   sdkInit: boolean;
 }
 interface ToolProps {
-  sdk: SendbirdGroupChat;
+  sdk: SdkStore['sdk'];
   logger: Logger;
-  messageSearchDispatcher: ({ type: string, payload: any }) => void;
+  messageSearchDispatcher: (param: { type: string, payload: any }) => void;
```
liamcho, line 13:
```suggestion
  messageSearchDispatcher: (props: { type: string, payload: any }) => void;
```

src/modules/MessageSearch/context/hooks/useSearchStringEffect.ts
```diff
@@ -6,7 +6,7 @@ interface DynamicParams {
 }
 
 interface StaticParams {
-  messageSearchDispatcher: ({ type: string, payload: any }) => void;
+  messageSearchDispatcher: (param: { type: string, payload: any }) => void;
```
liamcho, line 9:
```suggestion
  messageSearchDispatcher: (props: { type: string, payload: any }) => void;
```

src/modules/OpenChannel/components/OpenChannelMessage/index.tsx
```diff
@@ -114,116 +99,130 @@ export default function MessagOpenChannelMessageeHoc(props: OpenChannelMessagePr
...
       />
     );
   }
 
   return (
     <div className="sendbird-msg-hoc sendbird-msg--scroll-ref">
-      {/* date-separator */}
-      {
-        (hasSeparator && message?.createdAt) && (
+      <>
```
liamcho, line 111: why is this needed?

src/lib/selectors.ts
bang9: `SendBirdState` already has the type specified for the SDK.
During the returning it, the function's return type is correctly inferred

src/modules/OpenChannel/context/hooks/useResendMessageCallback.ts
```diff
@@ -21,65 +20,99 @@ function useResendMessageCallback(
...
-    } else {
-      // to alert user on console
-      // eslint-disable-next-line no-console
-      console.error('OpenChannel | useResendMessageCallback: Message is not resendable');
-      logger.warning('OpenChannel | useResendMessageCallback: Message is not resendable', failedMessage);
-    }
-  }, [currentOpenChannel]);
+    },
+    [currentOpenChannel],
+  );
```
liamcho, line 115: nice

src/ui/OpenchannelUserMessage/index.tsx
```diff
@@ -202,7 +202,7 @@ export default function OpenchannelUserMessage({
                   key={uuidv4()}
                   type={LabelTypography.BODY_1}
                   color={LabelColors.ONBACKGROUND_2}
-                  calssName="sendbird-openchannel-user-message-word"
+                  className="sendbird-openchannel-user-message-word"
```
liamcho, line 205: nice

tsconfig.json
```diff
@@ -2,10 +2,19 @@
   "compilerOptions": {
     "jsx": "react",
     "target": "es5",
+    "module": "es2020",
     "esModuleInterop": true,
+    "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
     "skipLibCheck": true,
-    "typeRoots": ["node_modules/@types", "node_modules/jest-extended", "types"]
+    "typeRoots": ["./node_modules/@types", "./types"]
```
liamcho, line 10: does this change matter?

scripts/post_build.js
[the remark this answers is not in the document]
liamcho: ah i see. Great, thanks!

src/lib/selectors.ts
liamcho: ok but personally i think its better to keep return type for all functions for consistency and readability. But this is minor.

babel.config.js
bang9: I added this to resolve the Storybook error.
It serves the purpose of modifying certain transpile options in Babel, resulting in some differences in the generated code.

But we did not use the 'loose' option, so I removed it, and when there are plugins included in preset-env but not explicitly added, it can cause these errors. I've found and applied these changes to resolve the issue.
https://github.com/babel/babel/issues/11622#issuecomment-638609015

66b488a

package.json
bang9: I don't see any significant changes that stand out!

scripts/post_build.js
bang9: In the build configuration (rollup.config.mjs), no files are generated under cjs/dist, except for CSS

src/lib/selectors.ts
bang9: Yes, it can vary based on personal preferences.
I personally don't prefer redundant type declarations for inferred types.
I tend to specify them only where it's necessary, as it can make the code harder to read when it gets longer.

If a function has multiple branches and complex logic with many possible return types, I do specify the return type for such cases to prevent errors

src/modules/Channel/components/TypingIndicator.tsx
bang9: I've changed this as a component (text -> component).

Are there any rules for type specification in components? It seems fragmented. It would be good to decide on this now.

const Component: React.FC<Props> = (props) => { ... };
const Component = (props: Props) => { ... } // Inference as JSX.Element;
const Component = (props: Props): JSX.Element => { ... }
const Component = (props: Props): ReactElement => { ... }
Personally, I prefer option 2.
By the way, JSX.Element is simply an interface that inherits from ReactElement (they are the same type)

src/modules/CreateChannel/components/InviteUsers/index.tsx
bang9: The userListQuery of InviteUsers is received from CreateChannelProvider. 
image

So, I've changed type to the same type as the one declared internally, which is UserListQuery.
image

src/modules/MessageSearch/context/hooks/useSetChannel.ts
bang9: In a typical development environment, we might use 'props,' but in the context of developing with React, 'props' feels like a reserved word primarily used within components. Therefore, I've used 'params

src/modules/MessageSearch/context/hooks/useSearchStringEffect.ts
bang9: In a typical development environment, we might use 'props,' but in the context of developing with React, 'props' feels like a reserved word primarily used within components. Therefore, I've used 'params'

tsconfig.json
bang9: It won't make a big difference; just doing it for a bit more clarity 🌝

src/modules/OpenChannel/components/OpenChannelMessage/index.tsx
bang9: image
It's because of this error.
Perhaps the code is quite complex, and the compiler doesn't seem to recognize it correctly.
That's why I simply wrapped it with a Fragment

src/modules/Channel/components/TypingIndicator.tsx
AhyoungRyu: We don't have any. I prefer 2 too :)

src/modules/MessageSearch/context/hooks/useSearchStringEffect.ts
AhyoungRyu: Yeah I generally we use `props` for components and `params` for others including hooks like here.

HoonBaek (approved):
Super awesome job! Thank you @bang9 I really appreciate this!

liamcho (approved):
LGTM, awesome job @bang9 !

src/modules/MessageSearch/context/hooks/useSearchStringEffect.ts
liamcho: ah its minor. I just suggested because the other one is using props.

bang9:
I'll merge it after testing in the sample environment.

PR #822, opened 2023-10-26
fix: Remove a logic blocking adding an empty channel to the ChannelList

HoonBaek:
[CLNP-1196](https://sendbird.atlassian.net/browse/CLNP-1196)

### Issue
We have blocked adding a new empty channel to the ChannelList. But it's not considered with the ChannelListQuery, and there was a use case that needed to add super group channels to the ChannelList.

It's fine to remove this logic because ChannelListQuery filters the empty channels by referring to its `includeEmpty` options.

### Fix
* Remove the logic blocking adding an empty channel to the ChannelList because the ChannelListQuery should judge it

[CLNP-1196]: https://sendbird.atlassian.net/browse/CLNP-1196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/ChannelList/dux/reducers.ts
+++ b/src/modules/ChannelList/dux/reducers.ts
@@ -140,9 +140,6 @@ export default function channelListReducer(
           const { allChannels = [], currentUserId, currentChannel, channelListQuery, disableAutoSelect } = state;
           const { unreadMessageCount } = channel;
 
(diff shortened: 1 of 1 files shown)


liamcho (approved):
LGTM

PR #825, opened 2023-10-30
feat: add suggested replies

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-581 
The feature request is from AI chat bot. 

#### Scenarios 
If a trigger message(can be set from the dashboard) is sent to the channel from a user (e.g. `Show me suggested replies` in this case), the predefined suggested reply options will be displayed. 
Once the user clicks an option from the list, the option list will be gone and the message will be sent. 
record

### DIFF
--- a/src/modules/Channel/components/Message/index.tsx
+++ b/src/modules/Channel/components/Message/index.tsx
@@ -27,6 +27,7 @@ import { EveryMessage, RenderCustomSeparatorProps, RenderMessageProps } from '..
 import { useLocalization } from '../../../../lib/LocalizationContext';
 import { useHandleOnScrollCallback } from '../../../../hooks/useHandleOnScrollCallback';
 import { useDirtyGetMentions } from '../../../Message/hooks/useDirtyGetMentions';
+import SuggestedReplies from '../SuggestedReplies';
 
 type MessageUIProps = {
   message: EveryMessage;
@@ -91,6 +92,7 @@ const Message = ({
     onMessageHighlighted,
     onScrollCallback,
     setIsScrolled,
+    sendMessage,
   } = useChannelContext();
   const [showEdit, setShowEdit] = useState(false);
   const [showRemove, setShowRemove] = useState(false);
@@ -208,6 +210,10 @@ const Message = ({
     }
     return null;
   }, [message, renderCustomSeparator]);
+
+  const suggestedReplies = useMemo(() => {
+    return message.extendedMessagePayload?.suggested_replies as string[] | undefined ?? [];
+  }, [(message.extendedMessagePayload?.suggested_replies as string[] | undefined)?.length]);
 
   if (renderedMessage) {
     return (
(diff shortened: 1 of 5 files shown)


src/modules/Channel/components/Message/index.tsx
```diff
@@ -209,6 +211,10 @@ const Message = ({
     return null;
   }, [message, renderCustomSeparator]);
 
+  const suggestedReplies = useMemo(() => {
+    return message.extendedMessagePayload?.suggested_replies as string[] | undefined ?? [];
+  }, [(message.extendedMessagePayload?.suggested_replies as string[] | undefined)?.length]);
```
AhyoungRyu, line 216: The `extendedMessagePayload` will be included in the upcoming JS chat SDK. So I need to push a commit to bump up the chat sdk version once it's release.

src/modules/Channel/components/SuggestedReplies/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return (
+    <div className="sendbird-suggested-replies">
+      {replyOptions.map((option: string) => {
+        return (
+          <div
+            className="sendbird-suggested-replies__option"
+            id={option}
+            key={option}
```
bang9, line 32: If the value can be set by the customer on the dashboard, it may be a good idea to add an index because we can't be sure it's unique. `key={index+option}`

src/modules/Channel/components/SuggestedReplies/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+const SuggestedReplies = ({ replyOptions, onSendMessage }: Props) => {
+  const [replied, setReplied] = useState<boolean>(false);
+
+  const onClickReply = (
+    event: React.MouseEvent<HTMLDivElement>,
+    option: string,
+  ) => {
+    event.preventDefault();
+    onSendMessage({ message: option });
+    setReplied(true);
```
bang9, line 18: How do we know that this message has been responded to when re-entering the channel next time, once the `replied` state has been reset?

bang9: Ah, I checked here `message.messageId === currentGroupChannel?.lastMessage.messageId`
never mind!

src/modules/Channel/components/Message/index.tsx
bang9: If it's a payload that is included in a specific message, it seems like there won't be significant changes. Should we consider removing the useMemo?

bang9 (approved):
lgtm!

src/modules/Channel/components/Message/index.tsx
AhyoungRyu: Yeah it makes sense. Just addressed it https://github.com/sendbird/sendbird-uikit-react/pull/825/commits/13206b3337445d1f7c1673aac7be6542674ea808 :)

src/modules/Channel/components/SuggestedReplies/index.tsx
AhyoungRyu: Sure it's been addressed too https://github.com/sendbird/sendbird-uikit-react/pull/825/commits/13206b3337445d1f7c1673aac7be6542674ea808

PR #828, opened 2023-10-31
fix: lastMessage is nullable

bang9:
- I fixed the runtime error that was occurring in the empty channel.


AhyoungRyu:
Thanks Airen for fixing this! It was my bad 😅
Just pushed [a commit](https://github.com/sendbird/sendbird-uikit-react/pull/828/commits/703540143463cf96dc10790b8afd63fee4980081) to addressed your feedback in https://github.com/sendbird/chat-ai-widget/pull/65#discussion_r1377385529 :)

AhyoungRyu (approved):

PR #829, opened 2023-11-02
feat: export initialMessagesFetch callback from hook

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-690

I took the logic inside of useInitialMessagesFetch's effect hook and made a callback function. And made the function to be returned from the hook to make it be from outside of UIKit. 

#### Before
```
function useInitialMessagesFetch(params): void {
  useEffect(() => {
     all the logics..
  }, [deps]}
}
```

#### After
```
function useInitialMessagesFetch(params): () => void {
  const fetchMessages() => useCallback(() => {
    all the logics..
  , [deps]}
  useEffect(() => {
     fetchMessages();
  }, [fetchMessages]}
  
  return fetchMessages;
}
```

### DIFF
--- a/src/modules/Channel/context/hooks/useInitialMessagesFetch.ts
+++ b/src/modules/Channel/context/hooks/useInitialMessagesFetch.ts
@@ -37,6 +37,13 @@ function useInitialMessagesFetch(
 ): () => void {
   const channelUrl = currentGroupChannel?.url;
 
+  /**
+   * useCallback(() => {}, [currentGroupChannel]) was buggy, that is why we did
+   * const channelUrl = currentGroupChannel && currentGroupChannel.url;
+   * useCallback(() => {}, [channelUrl])
+   * Again, this hook is supposed to execute when currentGroupChannel changes
+   * The 'channelUrl' here is not the same memory reference from Conversation.props
+   */
   const fetchMessages = useCallback(() => {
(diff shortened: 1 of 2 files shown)


src/modules/Channel/context/hooks/useInitialMessagesFetch.ts
```diff
@@ -129,6 +129,12 @@ function useInitialMessagesFetch(
         });
     }
   }, [channelUrl, userFilledMessageListQuery, initialTimeStamp]);
```
bang9, line 131: Can we update `useEffect` to `useCallback`  and move it under the useCallback deps? (OR remove)

```
  /**
   * Note - useEffect(() => {}, [currentGroupChannel])
   * was buggy, that is why we did
   * const channelUrl = currentGroupChannel && currentGroupChannel.url;
   * useEffect(() => {}, [channelUrl])
   * Again, this hook is supposed to execute when currentGroupChannel changes
   * The 'channelUrl' here is not the same memory reference from Conversation.props
   */
```

AhyoungRyu: Ahh yeah couldn't even realized that the note was there 😅 I think it's better to keep just in case. c69a1a608b017b8eaf98731d5bafbd9c274d3fc8

PR #830, opened 2023-11-03
bug-fix: Use channel members instead of fetch

liamcho:
Fixes: [CLNP-1099](https://sendbird.atlassian.net/browse/CLNP-1099)

[CLNP-1099]: https://sendbird.atlassian.net/browse/CLNP-1099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/ChannelSettings/components/ModerationPanel/MemberList.tsx
+++ b/src/modules/ChannelSettings/components/ModerationPanel/MemberList.tsx
@@ -1,27 +1,26 @@
-import React, {
-  ReactElement,
-  useEffect,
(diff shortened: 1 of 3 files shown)


bang9:
It looks like I've made corrections to the incorrect parts. We need to modify the MentionSuggestionList.

liamcho:
We should make new PR that applies this logic in MentionSuggestion part only

PR #831, opened 2023-11-03
fix: precise object dependency in effect hook

AhyoungRyu:
Found the issue while I was testing https://sendbird.atlassian.net/browse/SBISSUE-14025.

If user passes custom `queries(: object)` prop to Channel module, it'll cause unnecessary re-rendering in this effect hook. So I just made it more precise so the deps array don;t waste the resource to compare the object diff and make unnecessary side effect.


liamcho (approved):
LGTM!

PR #832, opened 2023-11-03
bug-fix: Use channel members instead of fetch when searched.

liamcho:
Fixes: [CLNP-1099](https://sendbird.atlassian.net/browse/CLNP-1099)

[CLNP-1099]: https://sendbird.atlassian.net/browse/CLNP-1099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Channel/components/SuggestedMentionList/index.tsx
+++ b/src/modules/Channel/components/SuggestedMentionList/index.tsx
@@ -69,6 +69,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
     ).slice(0, maxSuggestionCount);
   }, [
     channelInstance.members.length,
+    channelInstance.members.map((member: Member) => member.nickname).join(),
     channelInstance.members.map((member: Member) => member.state).join(),
   ])
 
@@ -130,6 +131,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
     channelInstance?.url,
     channelInstance.members.length,
     // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.nickname).join(),
     channelInstance.members.map((member: Member) => member.state).join(),
     searchString,
   ]);
--- a/src/modules/Channel/components/SuggestedMentionList/index.tsx
+++ b/src/modules/Channel/components/SuggestedMentionList/index.tsx
@@ -58,6 +58,13 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
   const [currentUser, setCurrentUser] = useState<User>(null);
   const [currentMemberList, setCurrentMemberList] = useState<Member[]>([]);
 
+  // We have to be specific like this or React would not recognize the changes in instances.
+  const currentSearchedMemberList = [
+    channelInstance.members.map((member: Member) => member.userId).join(),
+    channelInstance.members.map((member: Member) => member.nickname).join(),
+    channelInstance.members.map((member: Member) => member.isActive).join(),
+  ];
+
   const getCurrentSearchedMemberList = (): Member[] => useMemo(() => {
     return currentChannel.members
       .sort((a, b) => a.nickname?.localeCompare(b.nickname))
@@ -66,11 +73,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
         && member.userId !== currentUser?.userId
         && member.isActive,
       ).slice(0, maxSuggestionCount);
-  }, [
-    channelInstance.members.length,
-    channelInstance.members.map((member: Member) => member.nickname).join(),
-    channelInstance.members.map((member: Member) => member.isActive).join(),
-  ]);
+  }, currentSearchedMemberList);
 
   useEffect(() => {
     clearTimeout(timer);
@@ -128,11 +131,8 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
     }
   }, [
     channelInstance?.url,
-    channelInstance.members.length,
-    // We have to be specific like this or React would not recognize the changes in instances.
-    channelInstance.members.map((member: Member) => member.nickname).join(),
-    channelInstance.members.map((member: Member) => member.isActive).join(),
     searchString,
+    ...currentSearchedMemberList,
   ]);
 
   /* Fetch member list for super channel */
--- a/src/modules/Channel/components/SuggestedMentionList/index.tsx
+++ b/src/modules/Channel/components/SuggestedMentionList/index.tsx
@@ -13,7 +13,6 @@ import { MAX_USER_MENTION_COUNT, MAX_USER_SUGGESTION_COUNT, USER_MENTION_TEMP_CH
 import { MessageInputKeys } from '../../../../ui/MessageInput/const';
 import uuidv4 from '../../../../utils/uuid';
 import { useThreadContext } from '../../../Thread/context/ThreadProvider';
-import {MemberState} from '@sendbird/chat/groupChannel';
 
 export interface SuggestedMentionListProps {
   className?: string;
@@ -61,13 +60,12 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
 
   function getCurrentSearchedMemberList(): Member[] {
     return currentChannel.members
-    .sort((a, b) => a.nickname?.localeCompare(b.nickname))
-    .filter(
-      (member) =>
-        member.nickname?.toLowerCase().startsWith(searchString.toLowerCase()) &&
-        member.userId !== currentUser?.userId &&
-        member.isActive,
-    ).slice(0, maxSuggestionCount);
+      .sort((a, b) => a.nickname?.localeCompare(b.nickname))
+      .filter(
+        (member) => member.nickname?.toLowerCase().startsWith(searchString.toLowerCase())
+        && member.userId !== currentUser?.userId
+        && member.isActive,
+      ).slice(0, maxSuggestionCount);
   }
 
   useEffect(() => {
@@ -123,25 +121,25 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
         nicknameStartsWithFilter: searchString.slice(USER_MENTION_TEMP_CHAR.length),
       });
       query.next()
-      .then((memberList) => {
-        const suggestingMembers = memberList
-        .filter((member) => currentUserId !== member?.userId)
-        .slice(0, maxSuggestionCount);
-        if (suggestingMembers.length < 1) {
-          logger.info('SuggestedMentionList: Fetched member list is empty');
-        } else {
-          logger.info('SuggestedMentionList: Fetching member list succeeded', { memberListQuery: query, memberList: suggestingMembers });
-          setCurrentUser(suggestingMembers[0]);
-        }
-        setLastSearchString(searchString);
-        onFetchUsers(suggestingMembers);
-        setCurrentMemberList(suggestingMembers);
-      })
-      .catch((error) => {
-        if (error) {
-          logger.error('SuggestedMentionList: Fetching member list failed', error);
-        }
-      });
+        .then((memberList) => {
+          const suggestingMembers = memberList
+            .filter((member) => currentUserId !== member?.userId)
+            .slice(0, maxSuggestionCount);
+          if (suggestingMembers.length < 1) {
+            logger.info('SuggestedMentionList: Fetched member list is empty');
+          } else {
+            logger.info('SuggestedMentionList: Fetching member list succeeded', { memberListQuery: query, memberList: suggestingMembers });
+            setCurrentUser(suggestingMembers[0]);
+          }
+          setLastSearchString(searchString);
+          onFetchUsers(suggestingMembers);
+          setCurrentMemberList(suggestingMembers);
+        })
+        .catch((error) => {
+          if (error) {
+            logger.error('SuggestedMentionList: Fetching member list failed', error);
+          }
+        });
     } else {
       const suggestingMembers = getCurrentSearchedMemberList();
       if (suggestingMembers.length < 1) {
@@ -159,7 +157,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
     channelInstance.members.length,
     // We have to be specific like this or React would not recognize the changes in instances.
     channelInstance.members.map((member: Member) => member.state).join(),
-    searchString
+    searchString,
   ]);
 
   if (!ableAddMention && currentMemberList.length === 0) {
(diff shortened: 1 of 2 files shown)


src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -129,7 +142,25 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      }
+      setLastSearchString(searchString);
+      onFetchUsers(suggestingMembers);
+      setCurrentMemberList(suggestingMembers);
+    }
+  }, [
+    channelInstance?.url,
+    channelInstance.members.length,
+    // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.state).join(),
```
liamcho, line 161: @AhyoungRyu @bang9 please check this part

HoonBaek (approved):
LGTM
However, make sure checking if it works well in the storybook before merging

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -13,6 +13,7 @@ import { MAX_USER_MENTION_COUNT, MAX_USER_SUGGESTION_COUNT, USER_MENTION_TEMP_CH
 import { MessageInputKeys } from '../../../../ui/MessageInput/const';
 import uuidv4 from '../../../../utils/uuid';
 import { useThreadContext } from '../../../Thread/context/ThreadProvider';
+import {MemberState} from '@sendbird/chat/groupChannel';
```
HoonBaek, line 16: maybe space is required here?

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +56,17 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+
+  function getCurrentSearchedMemberList(): Member[] {
+    return currentChannel.members
+      .sort((a, b) => a.nickname?.localeCompare(b.nickname))
+      .filter(
+        (member) => member.nickname?.toLowerCase().startsWith(searchString.toLowerCase())
+        && member.userId !== currentUser?.userId
+        && member.isActive,
+      ).slice(0, maxSuggestionCount);
+  }
```
AhyoungRyu, line 69: Since it doesn't get a parameter and manipulate the param data, probably it's better to define as const \w useMemo(). 
```suggestion
  const currentSearchedMemberList(): Member[] => useMemo(() => {
    return currentChannel.members
    .sort((a, b) => a.nickname?.localeCompare(b.nickname))
    .filter(
      (member) =>
        member.nickname?.toLowerCase().startsWith(searchString.toLowerCase()) &&
        member.userId !== currentUser?.userId &&
        member.isActive,
    ).slice(0, maxSuggestionCount);
  }, [currentChannel.members.length])
```

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +56,17 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
   const [searchString, setSearchString] = useState('');
   const [lastSearchString, setLastSearchString] = useState('');
   const [currentUser, setCurrentUser] = useState<User>(null);
-  const [currentMemberList, setCurrentMemberList] = useState<Array<Member>>([]);
+  const [currentMemberList, setCurrentMemberList] = useState<Member[]>([]);
+
+  function getCurrentSearchedMemberList(): Member[] {
```
liamcho, line 61: does this have to useCallback @HoonBaek ?

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -104,32 +114,51 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      onFetchUsers(suggestingMembers);
+      setCurrentMemberList(suggestingMembers);
+    }
+  }, [
+    channelInstance?.url,
+    channelInstance.members.length,
+    // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.state).join(),
+    searchString,
+  ]);
```
HoonBaek, line 161: I think we can divide this useEffect into two
one is using the query, and the one is using members

src/modules/Channel/components/SuggestedMentionList/index.tsx
HoonBaek: Yeah, good to have

AhyoungRyu: Oh I just left a comment here https://github.com/sendbird/sendbird-uikit-react/pull/832#discussion_r1381209395 about this.

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +56,18 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+  const currentSearchedMemberList(): Member[] => useMemo(() => {
+    return currentChannel.members
+    .sort((a, b) => a.nickname?.localeCompare(b.nickname))
+    .filter(
+      (member) =>
+        member.nickname?.toLowerCase().startsWith(searchString.toLowerCase()) &&
+        member.userId !== currentUser?.userId &&
+        member.isActive,
+    ).slice(0, maxSuggestionCount);
+  }, [currentChannel.members.length])
```
HoonBaek, line 70: ```currentChannel.members.map(member => member.nickname)```
will be better, I guess

HoonBaek: oh, it's my mistake, the other properties also being used
I think `currentChannel.members.length` is enough then... sorry;; @liamcho

HoonBaek: If you want to make it tight
```
currentChannel.members.map(member => member.nickname),
currentChannel.members.map(member => member.userId),
currentChannel.members.map(member => member.isActive),
```

This will work, It seems `member.state` is not used here

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -93,43 +108,75 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      setLastSearchString(searchString);
+      onFetchUsers(suggestingMembers);
+      setCurrentMemberList(suggestingMembers);
+    }
+  }, [
+    channelInstance?.url,
+    channelInstance.members.length,
+    // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.nickname).join(),
+    channelInstance.members.map((member: Member) => member.state).join(),
```
HoonBaek: You can define the CurrentSearchedMemberList above and put it here instead of these

src/modules/Channel/components/SuggestedMentionList/index.tsx
liamcho: I think length will do for userId one. I changed to isActive!

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -93,43 +107,75 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      setLastSearchString(searchString);
+      onFetchUsers(suggestingMembers);
+      setCurrentMemberList(suggestingMembers);
+    }
+  }, [
+    channelInstance?.url,
+    channelInstance.members.length,
+    // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.nickname).join(),
+    channelInstance.members.map((member: Member) => member.isActive).join(),
```
HoonBaek, line 134: you can define the currentSearchedMemberList above and put it here instead of these

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +56,21 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      .sort((a, b) => a.nickname?.localeCompare(b.nickname))
+      .filter(
+        (member) => member.nickname?.toLowerCase().startsWith(searchString.toLowerCase())
+        && member.userId !== currentUser?.userId
+        && member.isActive,
+      ).slice(0, maxSuggestionCount);
+  }, [
+    channelInstance.members.length,
+    channelInstance.members.map((member: Member) => member.nickname).join(),
+    channelInstance.members.map((member: Member) => member.isActive).join(),
```
HoonBaek, line 72:
```suggestion
    channelInstance.members.map((member: Member) => member.nickname).join(),
    channelInstance.members.map((member: Member) => member.userId).join(),
    channelInstance.members.map((member: Member) => member.isActive).join(),
```

HoonBaek (approved):

src/modules/Channel/components/SuggestedMentionList/index.tsx
bang9: I believe it's not a good idea to separate the same business logic because it would require managing two useEffect blocks.
Simply by determining where to fetch the user based on isSuper, you can decide where to handle user retrieval. Instead of separating the useEffect itself, it would be more beneficial to split the function that retrieves members into two, reducing the complexity of the useEffect and making the code more readable.

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +56,24 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+
+  const getCurrentSearchedMemberList = (): Member[] => useMemo(() => {
+    return currentChannel.members
+      .sort((a, b) => a.nickname?.localeCompare(b.nickname))
+      .filter(
+        (member) => member.nickname?.toLowerCase().startsWith(searchString.toLowerCase())
+        && member.userId !== currentUser?.userId
+        && member.isActive,
+      ).slice(0, maxSuggestionCount);
+  }, currentSearchedMemberList);
```
bang9, line 76: It seems like calculations and memoization are being performed even when they are not necessary.
It appears to be a better approach to create a separate function and generate the list only when needed by calling it.

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -93,43 +110,72 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
...
+      const query = channelInstance?.createMemberListQuery({
+        limit: maxSuggestionCount + 1, // because current user could be included
+        nicknameStartsWithFilter: searchString.slice(USER_MENTION_TEMP_CHAR.length),
       });
-  }, [channelInstance?.url, searchString]);
+      query.next()
+        .then((memberList) => {
+          const suggestingMembers = memberList
+            .filter((member) => currentUserId !== member?.userId)
+            .slice(0, maxSuggestionCount);
```
bang9, line 158: How about creating a function to handle and manage the part that retrieves users?

```ts
async function fetchMembersFromChannel(channel, maxSuggestionCount, searchString): Promise<Member[]> {
    return channel.members
      .sort((a, b) => a.nickname?.localeCompare(b.nickname))
      .filter(
        (member) => member.nickname?.toLowerCase().startsWith(searchString.toLowerCase())
        && member.userId !== currentUser?.userId
        && member.isActive,
      ).slice(0, maxSuggestionCount);
}

async function fetchMembersFromQuery(channel, maxSuggestionCount, searchString): Promise<Member[]> {
    const query = channel.createMemberListQuery({
      limit: maxSuggestionCount + 1, // because current user could be included
      nicknameStartsWithFilter: searchString.slice(USER_MENTION_TEMP_CHAR.length),
    });

    return query.next()
      .then((memberList) => {
        return memberList
          .filter((member) => currentUserId !== member?.userId)
          .slice(0, maxSuggestionCount);
      }
}
```

src/modules/Channel/components/SuggestedMentionList/index.tsx
liamcho: Can we achieve that by replacing useMemo with useCallback?

bang9: It looks like it can be made into a pure function.

liamcho: I see. ill put back to pure function

src/modules/Channel/components/SuggestedMentionList/index.tsx
liamcho: Thanks for the great suggestion. I have applied this change but the dependency I used is a bit different.
```[
    channelInstance?.url,
    // We have to be specific like this or React would not recognize the changes in instances.
    channelInstance.members.map((member: Member) => member.userId).join(),
    channelInstance.members.map((member: Member) => member.nickname).join(),
    channelInstance.members.map((member: Member) => member.isActive).join(),
    searchString,
    maxSuggestionCount,
    currentUserId,
    currentMemberList,
    lastSearchString
  ]
```
let me know if this looks ok for you.

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -56,7 +57,7 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
   const [searchString, setSearchString] = useState('');
   const [lastSearchString, setLastSearchString] = useState('');
   const [currentUser, setCurrentUser] = useState<User>(null);
```
bang9, line 59: nit:
Can we change the variable name of **currentUser** even though it's unrelated to this PR's changes? (e.g., to currentFocusedMember)
Normally, in our business logic, **currentUser** refers to the user currently connected to the SDK, so the name can be quite confusing.

src/modules/Channel/components/SuggestedMentionList/index.tsx
```diff
@@ -129,7 +123,18 @@ function SuggestedMentionList(props: SuggestedMentionListProps): JSX.Element {
           logger.error('SuggestedMentionList: Fetching member list failed', error);
         }
       });
-  }, [channelInstance?.url, searchString]);
+  }, [
+    channelInstance?.url,
+    // We have to be specific like this or React would not recognize the changes in instances.
+    channelInstance.members.map((member: Member) => member.userId).join(),
```
bang9, line 129: Since user ids are not expected to change within the same member list, so it seems like this line can be removed.

src/modules/Channel/components/SuggestedMentionList/index.tsx
liamcho: ah thanks for finding this out. ill change to currentFocusedMember

src/modules/Channel/components/SuggestedMentionList/index.tsx
liamcho: that was astute. i remove it!

bang9 (approved):

PR #839, opened 2023-11-09
fix: Improve Image Compression feature

HoonBaek:
[CLNP-1320](https://sendbird.atlassian.net/browse/CLNP-1320)

### Fixes
* Use the more impactful value between the `resizingWidth` and `resizingHeight`
  * So, the original images' ratio won't be broken
* Apply the ImageCompression to the Thread module
* Apply the ImageCompression for sending `file message` and `multiple files message`

### Fixes
* Create a `compressImages` util function
* Compress the images in the `useHandleUploadFiles` hook
  * Do not compress the image inside of the `useSendFileMessageCallback` hook of the Channel module

[CLNP-1320]: https://sendbird.atlassian.net/browse/CLNP-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
+++ b/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
@@ -11,7 +11,7 @@ import { ButtonTypes } from '../../../../ui/Button';
 import { useLocalization } from '../../../../lib/LocalizationContext';
 import { ModalFooter } from '../../../../ui/Modal';
 import { FileMessage, MultipleFilesMessage } from '@sendbird/chat/message';
-import { compressImages } from '../../../../utils/compressImage';
+import { compressImages } from '../../../../utils/compressImages';
 
 /**
  * The handleUploadFiles is a function sending a FileMessage and MultipleFilesMessage
--- a/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
+++ b/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
@@ -94,12 +94,12 @@ export const useHandleUploadFiles = ({
     }
 
     // Image Compression
-    const { compressedImages, notCompressedFiles } = await compressImages({
+    const { compressedFiles } = await compressImages({
       files,
       imageCompression,
       logger,
     });
-    const sendingFiles = compressedImages.concat(notCompressedFiles);
+    const sendingFiles = compressedFiles;
 
     // Send File Message
     if (sendingFiles.length === 1) {
--- a/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
+++ b/src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
@@ -11,6 +11,7 @@ import { ButtonTypes } from '../../../../ui/Button';
 import { useLocalization } from '../../../../lib/LocalizationContext';
 import { ModalFooter } from '../../../../ui/Modal';
 import { FileMessage, MultipleFilesMessage } from '@sendbird/chat/message';
+import { compressImages } from '../../../../utils/compressImage';
 
 /**
  * The handleUploadFiles is a function sending a FileMessage and MultipleFilesMessage
@@ -25,7 +26,6 @@ interface useHandleUploadFilesDynamicProps {
 interface useHandleUploadFilesStaticProps {
   logger: Logger;
 }
-export type HandleUploadFunctionType = (files: File[]) => void;
 
 export const useHandleUploadFiles = ({
   sendFileMessage,
@@ -33,15 +33,16 @@ export const useHandleUploadFiles = ({
   quoteMessage,
 }: useHandleUploadFilesDynamicProps, {
   logger,
-}: useHandleUploadFilesStaticProps): Array<HandleUploadFunctionType> => {
+}: useHandleUploadFilesStaticProps) => {
   const { stringSet } = useLocalization();
   const { config } = useSendbirdStateContext();
+  const { imageCompression } = config;
   const uikitUploadSizeLimit = config?.uikitUploadSizeLimit;
   const uikitMultipleFilesMessageLimit = config?.uikitMultipleFilesMessageLimit;
   const { openModal } = useGlobalModalContext();
 
-  const handleUploadFiles = useCallback((files: File[]) => {
-    // Validate Paremeters
+  const handleUploadFiles = useCallback(async (files: File[]) => {
+    // Validate Parameters
     if (!sendFileMessage || !sendMultipleFilesMessage) {
       logger.warning('Channel|useHandleUploadFiles: required functions are undefined', { sendFileMessage, sendMultipleFilesMessage });
       return;
@@ -69,6 +70,7 @@ export const useHandleUploadFiles = ({
       });
       return;
     }
+    // Validate file sizes
     if (files.some((file: File) => file.size > uikitUploadSizeLimit)) {
       // The default value of uikitUploadSizeLimit is 26MB
       logger.info(`Channel|useHandleUploadFiles: Cannot upload file size exceeding ${uikitUploadSizeLimit}`);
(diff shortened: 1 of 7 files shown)


src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
...
+    logger.warning('useImageCompression: There is no file.', files);
+    return;
+  }
+  if (compressionRate < 0 || 1 < compressionRate) {
+    logger.warning('useImageCompression: The compressionRate is not acceptable.', compressionRate);
+    return;
+  }
+
+  const compressedImages: File[] = [];
+  const notCompressedFiles: File[] = [];
```
AhyoungRyu, line 92: Do we have a reason not to keep consistency in the terms `images` <-> `files`? If not, why don't we use simpler name here; `succeed & failed`
```suggestion
  const succeedFiles: File[] = [];
  const failedFiles: File[] = [];
```

src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
...
+  files,
+  logger,
+  imageCompression,
+}: CompressImagesProps) => {
+  const { compressionRate } = imageCompression;
+  const resizingWidth = pxToNumber(imageCompression.resizingWidth);
+  const resizingHeight = pxToNumber(imageCompression.resizingHeight);
+
+  if (!(Array.isArray(files) && files.length > 0)) {
+    logger.warning('useImageCompression: There is no file.', files);
```
AhyoungRyu, line 83: nit;
```suggestion
    logger.warning('useImageCompression: There is no files.', files);
```

src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
+import { ImageCompressionOptions } from '../lib/Sendbird';
+import pxToNumber from './pxToNumber';
+import { Logger } from '../lib/SendbirdState';
+
+interface CompressImageProps {
```
bang9, line 5:
```suggestion
interface CompressImageParams {
```

src/utils/compressImages.ts
bang9, line 83: wrong log prefix(`useImageCompression`)

src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
...
+  const resizingWidth = pxToNumber(imageCompression.resizingWidth);
+  const resizingHeight = pxToNumber(imageCompression.resizingHeight);
+
+  if (!(Array.isArray(files) && files.length > 0)) {
+    logger.warning('useImageCompression: There is no file.', files);
+    return;
+  }
+  if (compressionRate < 0 || 1 < compressionRate) {
+    logger.warning('useImageCompression: The compressionRate is not acceptable.', compressionRate);
+    return;
```
bang9, line 88: It would be good if the return type is consistent.

src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
```diff
@@ -91,15 +93,26 @@ export const useHandleUploadFiles = ({
...
     }
 
-    if (files.length === 1) {
+    // Image Compression
+    const { compressedImages, notCompressedFiles } = await compressImages({
+      files,
+      imageCompression,
+      logger,
+    });
+    const sendingFiles = compressedImages.concat(notCompressedFiles);
```
liamcho, line 102: Would this not change the image order? We should keep the order of selected files.

src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
...
+          }
+        },
+        imageFile.type,
+        compressionRate,
+      );
+    };
+  });
+};
+
+export interface CompressImagesProps {
```
bang9, line 68:
```suggestion
export interface CompressImagesParams {
```

src/utils/compressImages.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  if (compressionRate < 0 || 1 < compressionRate) {
+    logger.warning('useImageCompression: The compressionRate is not acceptable.', compressionRate);
+    return;
+  }
+
+  const compressedImages: File[] = [];
+  const notCompressedFiles: File[] = [];
+
+  await Promise.all(
```
liamcho, line 94: hmm compression could take a long time no? if we selected ten 25mb pdf files, it takes long time before actual send. However, I think there is no better option unless we keep our current sending logic that sends one file at a time not altogether at the same time.

liamcho (approved):
LGTM just check my comment for maintaining the order of selected files

src/utils/compressImages.ts
liamcho: minor: "There are no files."

src/modules/Channel/components/MessageInput/useHandleUploadFiles.tsx
HoonBaek: What a good catch! Thanks for catching this! Let me fix it

src/utils/compressImages.ts
HoonBaek: What is wrong?

HoonBaek: aha~! I got it

HoonBaek: Thank you for catching!

src/utils/compressImages.ts
HoonBaek: I changed the return values, pls check it again

src/utils/compressImages.ts
HoonBaek: hm... yes you are right;;
We are adding pending messages in the `onPending` handler
so it's impossible to add pending messages before compressing images for now..

src/modules/Channel/context/hooks/useSendFileMessageCallback.ts
```diff
@@ -28,169 +28,60 @@ export default function useSendFileMessageCallback(
...
-        };
+    (compressedFile: File, quoteMessage = null) => new Promise<FileMessage>((resolve, reject) => {
+      // Create FileMessageParams
+      let params: FileMessageCreateParams = onBeforeSendFileMessage?.(compressedFile, quoteMessage);
+      if (!params) {
+        params = { file: compressedFile };
         if (quoteMessage) {
           params.isReplyToChannel = true;
           params.parentMessageId = quoteMessage.messageId;
         }
```
bang9, line 39: Q. Why doesn't the params created for onBeforeSendFileMessage handle the following processing?
```
if (quoteMessage) {
  params.isReplyToChannel = true;
  params.parentMessageId = quoteMessage.messageId;
}
```

src/utils/compressImages.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      let targetResizingWidth = resizingWidth || originWidth;
+      let targetResizingHeight = resizingHeight || originHeight;
+
+      // Use the more impactful value, so the original images' ratio won't be broken
+      if (widthRatio > heightRatio) {
+        targetResizingHeight = originHeight / (resizingWidth ? widthRatio : 1);
+      } else if (heightRatio > widthRatio) {
+        targetResizingWidth = originWidth / (resizingHeight ? heightRatio : 1);
+      }
```
bang9, line 39: It would be good to extract the part that calculates the size into a function and write tests if possible.

src/utils/compressImages.ts
bang9: To determine if there is a significant impact in practice, it would be a good idea to conduct tests to see how long it takes to compress a 25MB image.

Promise.all processing is influenced by the time taken for the longest operation, so if the individual compression time is not too long, there shouldn't be a significant issue.

src/modules/Channel/context/hooks/useSendFileMessageCallback.ts
HoonBaek: It was not applying this process to the onBeforeSendFileMessage before.
The onBeforeSendFileMessage is a function that totally depends on the customer's input

src/utils/compressImages.ts
AhyoungRyu: Yeah right I missed that. Thanks Liam!

PR #848, opened 2023-11-16
bug-fix: Added and applied useOnScrollReachedEndDetector() in MessageList to mark new messages as read when scroll reaches bottom.

liamcho:
Fixes: [CLNP-1524](https://sendbird.atlassian.net/browse/CLNP-1524)

### Changelog
- Added and applied `useOnScrollReachedEndDetector()` in `MessageList` to mark new messages as read when scroll reaches bottom
- Channel list no longer displays unread message count for focused channel

The second change was due to limitation where isScrollBottom and hasNext states needs to be added globally but they are from channel context so channel list cannot see them with the current architecture.

[CLNP-1524]: https://sendbird.atlassian.net/browse/CLNP-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useOnScrollReachedEndDetector/index.ts
+++ b/src/hooks/useOnScrollReachedEndDetector/index.ts
@@ -7,28 +7,27 @@ import { usePreservedCallback } from '@sendbird/uikit-tools';
 const BUFFER_DELAY = 500;
 
 export interface UseOnScrollReachedEndDetectorProps {
-  scrollRef: React.RefObject<HTMLDivElement>;
   onReachedTop?: () => void;
   onReachedBottom?: () => void;
   onInBetween?: () => void;
 }
 
-export function useOnScrollPositionChangeDetector(props: UseOnScrollReachedEndDetectorProps): () => void {
+export function useOnScrollPositionChangeDetector(
+  props: UseOnScrollReachedEndDetectorProps,
+): (event: React.UIEvent<HTMLDivElement, UIEvent>) => void {
   const {
-    scrollRef,
     onReachedTop,
     onReachedBottom,
     onInBetween,
   } = props;
 
-  const cb = usePreservedCallback(() => {
-    const current = scrollRef?.current;
-    if (current) {
+  const cb = usePreservedCallback((event: React.UIEvent<HTMLDivElement, UIEvent>) => {
+    if (event?.target) {
       const {
         scrollTop,
         scrollHeight,
         clientHeight,
-      } = current;
+      } = event.target as HTMLDivElement;
       if (isAboutSame(scrollTop, 0, SCROLL_BUFFER)) {
         onReachedTop();
       } else if (isAboutSame(scrollHeight, clientHeight + scrollTop, SCROLL_BUFFER)) {
--- a/src/hooks/useOnScrollReachedEndDetector/index.ts
+++ b/src/hooks/useOnScrollReachedEndDetector/index.ts
@@ -1,4 +1,4 @@
-import React  from 'react';
+import React from 'react';
 import { SCROLL_BUFFER } from '../../utils/consts';
 import { isAboutSame } from '../../modules/Channel/context/utils';
 import { useDebounce } from '../useDebounce';
--- a/src/hooks/useOnScrollReachedEndDetector/index.ts
+++ b/src/hooks/useOnScrollReachedEndDetector/index.ts
@@ -1,0 +1,38 @@
+import React, { useCallback } from 'react';
+import { SCROLL_BUFFER } from '../../utils/consts';
+import { isAboutSame } from '../../modules/Channel/context/utils';
+import { useDebounce } from '../useDebounce';
+
+const BUFFER_DELAY = 500;
+
+export interface UseOnScrollReachedEndDetectorProps {
+  scrollRef: React.RefObject<HTMLDivElement>;
+  onReachedTop?: () => void;
+  onReachedBottom?: () => void;
+}
+
+export function useOnScrollReachedEndDetector(props: UseOnScrollReachedEndDetectorProps): () => void {
+  const {
+    scrollRef,
+    onReachedTop,
+    onReachedBottom,
+  } = props;
+
+  const cb = useCallback(() => {
+    const current = scrollRef?.current;
+    if (current) {
+      const {
+        scrollTop,
+        scrollHeight,
+        clientHeight,
+      } = current;
+      if (isAboutSame(scrollTop, 0, SCROLL_BUFFER)) {
+        onReachedTop();
+      } else if (isAboutSame(scrollHeight, clientHeight + scrollTop, SCROLL_BUFFER)) {
+        onReachedBottom();
+      }
+    }
+  }, [scrollRef, onReachedTop, onReachedBottom]);
+
+  return useDebounce(cb, BUFFER_DELAY);
+}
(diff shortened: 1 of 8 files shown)


src/hooks/useOnScrollReachedEndDetector/index.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface UseOnScrollReachedEndDetectorProps {
+  scrollRef: React.RefObject<HTMLDivElement>;
+  onReachedTop?: () => void;
+  onReachedBottom?: () => void;
+  onInBetween?: () => void;
+}
+
+export function useOnScrollPositionChangeDetector(props: UseOnScrollReachedEndDetectorProps): () => void {
+  const {
+    scrollRef,
```
bang9, line 17: Is it possible to use event from onScroll instead of scollRef?

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -245,29 +273,34 @@ const MessageList: React.FC<MessageListProps> = ({
...
+          (!isScrollBottom || hasMoreNext) && (unreadSince || unreadSinceDate) && (
+            <UnreadCount
+              className="sendbird-conversation__messages__notification"
+              count={currentGroupChannel?.unreadMessageCount}
+              time={unreadSince}
+              lastReadAt={unreadSinceDate}
+              onClick={() => {
+                if (scrollRef?.current?.scrollTop) {
+                  scrollRef.current.scrollTop = (scrollRef?.current?.scrollHeight ?? 0) - (scrollRef?.current?.offsetHeight ?? 0);
+                }
```
bang9, line 289: Is this logic for scrolling to the bottom?
```suggestion
                if (scrollRef?.current) {
                  scrollRef.current.scrollTop = Number.MAX_SAFE_INTEGER
                }
```

src/hooks/useOnScrollReachedEndDetector/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      } = current;
+      if (isAboutSame(scrollTop, 0, SCROLL_BUFFER)) {
+        onReachedTop();
+      } else if (isAboutSame(scrollHeight, clientHeight + scrollTop, SCROLL_BUFFER)) {
+        onReachedBottom();
+      } else {
+        onInBetween();
+      }
+    }
+  }, [scrollRef, onReachedTop, onReachedBottom, onInBetween]);
```
bang9, line 39:
```suggestion
import { usePreservedCallback } from "@sendbird/uikit-tools";


  const cb = usePreservedCallback(() => {
    const current = scrollRef?.current;
    if (current) {
      const {
        scrollTop,
        scrollHeight,
        clientHeight,
      } = current;
      if (isAboutSame(scrollTop, 0, SCROLL_BUFFER)) {
        onReachedTop();
      } else if (isAboutSame(scrollHeight, clientHeight + scrollTop, SCROLL_BUFFER)) {
        onReachedBottom();
      } else {
        onInBetween();
      }
    }
  });
```

src/hooks/useOnScrollReachedEndDetector/index.ts
liamcho: We could but im not sure how to change dependency from current:
```
[scrollRef, onReachedTop, onReachedBottom, onInBetween]
```
Could you enlight me please?

src/hooks/useOnScrollReachedEndDetector/index.ts
liamcho: what is usePreservedCallback?

src/modules/Channel/components/MessageList/index.tsx
liamcho: yeah it seems like it is hitting the bottom most. At first i thought this is going to top of the last message but when i tested it was going to the bottom of the last message. So i think we can change to your suggested.

src/hooks/useOnScrollReachedEndDetector/index.ts
bang9: Functions included in the dependency array must maintain their references, so they need to be wrapped with useCallback.
```tsx
const useMyHook = (callbackA, callbackB) => {
  return useCallback(() => {
    callbackA();
    callbackB();
  }, [callbackA, callbackB]);
}
```

When used like this, the use of useCallback is meaningless:
```tsx
const callback = useMyHook(() => { ... }, () => { ... });
```

Instead, you should use it like this, even though it makes the code complex and messy(useCallback hell):
```tsx
const callbackA = useCallback(() => { ... }, []);
const callbackB = useCallback(() => { ... }, []);
const callback = useMyHook(callbackA, callbackB);
```

By using usePreservedCallback, you can achieve the same result without worrying about these dependencies. The function references are always maintained consistently, and a new function is executed each time:
```tsx
const useMyHook = (callbackA, callbackB) => {
  return usePreservedCallback(() => {
    callbackA();
    callbackB();
  });
}

const callback = useMyHook(() => { ... }, () => { ... });
```

You can check the implementation here: https://github.com/sendbird/sendbird-uikit-core-ts/blob/main/packages/uikit-tools/src/hooks/usePreservedCallback.ts

liamcho: Thats pretty nice, thanks for this information. Ill apply this

src/modules/ChannelList/components/ChannelPreview/index.tsx
```diff
@@ -227,7 +228,12 @@ const ChannelPreview: React.FC<ChannelPreviewInterface> = ({
         https://github.com/facebook/react/issues/11387#issuecomment-340019419
       */}
       {
-        showMobileLeave && isMobile && (
+        /**
+         * Do not show unread count for focused channel. This is because of the limitation where
+         * isScrollBottom and hasNext states needs to be added globally but they are from channel context
+         * so channel list cannot see them with the current architecture.
+         */
+        (channel.url !== currentChannel?.url) && showMobileLeave && isMobile && (
```
HoonBaek, line 236: In my memory,
we call `markAsRead` when the message list scroll reaches the bottom,
and then the ChannelList receives the `onUnreadMemberCountUpdated` event,
so the unread count will be removed

**BTW, this is a channel leave button, not the UnreadCount badge**

src/hooks/useOnScrollReachedEndDetector/index.ts
bang9: If you switch to usePreservedCallback, it seems like the dependency array will no longer be necessary

src/modules/ChannelList/components/ChannelPreview/index.tsx
liamcho: The issue with that behaviour is blink like effect. When scroll is at the bottom,  unread count updates twice: 0 -> 1 -> 0. This is blink like effect and looks unnatural and customer complained about it. This is why we removed it.
It is UnreadCount badge. I tested? @HoonBaek

src/hooks/useOnScrollReachedEndDetector/index.ts
liamcho: ah then lets resolve this bc usePreservedCallback has been applied

bang9: Could we change to event instead of ref? Receiving values related to scrolling using a ref seems a bit awkward because it requires a separate execution of callbacks within the event handler.

To ensure consistency between the ref of the element and the actual element triggering the event, it appears more favorable to eliminate the ref and utilize the event directly.

liamcho: Done!

bang9 (approved):
lgtm!

PR #849, opened 2023-11-20
feat: Create a new GroupChannelList component

HoonBaek:
* Upgrade the importing `uikit-tools` version to `v0.0.1-alpha.43`
  for `useGroupChannelList` and `useGroupChannelEventHandler`
* Clone the ChannelList to GroupChannelList and Replace the dux to the `useGroupChannelList`
* Clean up the interface and props of the GroupChannelList


src/modules/App/types.ts
```diff
@@ -17,7 +17,7 @@ export interface AppLayoutProps {
   isMultipleFilesMessageEnabled?: boolean;
   allowProfileEdit?: boolean;
   showSearchIcon?: boolean;
-  onProfileEditSuccess?(user: User): void;
+  onProfileEditSuccess?(user: User): void; // TODO: Unused props, deprecate it
```
HoonBaek, line 20: TODO: Deprecate the onProfileEditSuccess (in App comp)

src/modules/CreateChannel/context/CreateChannelProvider.tsx
```diff
@@ -39,7 +39,12 @@ export interface CreateChannelContextInterface {
   createChannel: CreateChannel;
   sdk: SendbirdGroupChat;
   userListQuery?(): UserListQuery;
+  /**
+   * @deprecated
+   * Use the onClickCreateChannel instead
+   */
   overrideInviteUser?(params: OverrideInviteUserType): void;
+  // TODO: onClickCreateChannel instead of overrideInviteUser
```
HoonBaek, line 47: TODO: Rename the overrideInviteUser to onClickCreateChannel (in the CreateChannel)

HoonBaek (commented):
Deprecating Props

src/modules/CreateChannel/context/CreateChannelProvider.tsx
```diff
@@ -39,7 +39,12 @@ export interface CreateChannelContextInterface {
   createChannel: CreateChannel;
   sdk: SendbirdGroupChat;
   userListQuery?(): UserListQuery;
+  /**
+   * @deprecated
+   * Use the onClickCreateChannel instead
+   */
```
bang9, line 45: If you haven't actually deprecated it, it would be a good idea to add comments later when you actually work on it.

src/modules/GroupChannelList/components/ChannelListHeader/index.scss
bang9, line 1: If the CSS in the existing Channel module is exactly the same, it may be necessary to check for any potential side effects when building, as the namespace will be identical.
Otherwise, it might be a good idea to consider utilizing the CSS from the Channel module as is (import?), if it's different.

src/modules/GroupChannelList/components/ChannelListHeader/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+}: ChannelListHeaderInterface) => {
+  const { stores, config } = useSendbirdStateContext();
+  const { user } = stores.userStore;
+  const { logger } = config;
+
+  const { stringSet } = useContext(LocalizationContext);
+
+  if (renderHeader) {
+    logger?.warning('Recomend to use "renderTitle" instead of "renderHeader". It will be deprecated.');
+  }
```
bang9, line 33: If `renderHeader` has been deprecated in ChannelList, it seems safe to remove it in GroupChannelList.

HoonBaek: Good catch!

src/modules/GroupChannelList/components/ChannelListHeader/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    <div
+      className={[
+        'sendbird-channel-header',
+        allowProfileEdit ? 'sendbird-channel-header--allow-edit' : '',
+      ].join(' ')}
+    >
+      {
+        titleRenderer?.() || (
```
bang9, line 45: At the current moment, it's not particularly crucial, but, it seems necessary to discuss whether this rendering approach is suitable for the DX.
In React, typically, if a user does not want to render the component, they can explicitly specify null or undefined as the return value.
However, when using this approach, `Fragment` appears to be the only viable option.

```tsx
renderHeader={() => null} // not working
renderHeader={() => <React.Fragment />} // working
```

src/modules/GroupChannelList/components/ChannelListHeader/index.scss
HoonBaek: Yes, I didn't touch the CSS part yet
I will do it in the separated PR!

src/modules/GroupChannelList/components/ChannelListHeader/index.tsx
HoonBaek: That's a good idea, we can do that!

src/modules/GroupChannelList/components/AddChannel/index.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from 'react';
+
+import IconButton from '../../../../ui/IconButton';
+import Icon, { IconTypes, IconColors } from '../../../../ui/Icon';
+
+import CreateChannel from '../../../CreateChannel';
+import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
+import { useGroupChannelListContext } from '../../context/GroupChannelListProvider';
+
+export const AddChannel: React.VoidFunctionComponent = () => {
```
bang9, line 10:
```suggestion
export const AddChannel = () => {
```

src/modules/GroupChannelList/components/ChannelListHeader/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  onEdit?: (props: void) => void;
+  allowProfileEdit?: boolean;
+}
+
+const ChannelListHeader: React.FC<ChannelListHeaderInterface> = ({
+  renderTitle,
+  renderIconButton,
+  onEdit,
+  allowProfileEdit,
+}: ChannelListHeaderInterface) => {
```
bang9, line 22:
```suggestion
const ChannelListHeade = ({
  renderTitle,
  renderIconButton,
  onEdit,
  allowProfileEdit,
}: ChannelListHeaderInterface) => {
```

src/modules/GroupChannelList/components/ChannelListHeader/index.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useContext } from 'react';
+
+import { LocalizationContext } from '../../../../lib/LocalizationContext';
+import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
+import Label, { LabelTypography, LabelColors } from '../../../../ui/Label';
+
+import './index.scss';
+import Avatar from '../../../../ui/Avatar';
+
+interface ChannelListHeaderInterface {
```
bang9, line 10: Props?

src/modules/GroupChannelList/components/ChannelListUI/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+export interface ChannelListUIProps {
+  renderChannelPreview?: (props: RenderChannelPreviewProps) => React.ReactElement;
+  renderUserProfile?: (props: RenderUserProfileProps) => React.ReactElement;
+  renderHeader?: (props: void) => React.ReactElement;
+  renderPlaceHolderError?: (props: void) => React.ReactElement;
+  renderPlaceHolderLoading?: (props: void) => React.ReactElement;
+  renderPlaceHolderEmptyList?: (props: void) => React.ReactElement;
+}
+
+const ChannelListUI: React.FC<ChannelListUIProps> = (props: ChannelListUIProps) => {
```
bang9, line 41:
```suggestion
const ChannelListUI = (props: ChannelListUIProps) => {
```

src/modules/GroupChannelList/stories/index.stories.js
bang9, line 1: Let's remove stories 🌝

src/modules/GroupChannelList/components/Placeholder.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import Placeholder, { PlaceHolderTypes } from '../../../ui/PlaceHolder';
+
+type ChannelsPlaceholderProps = {
+  type: keyof typeof PlaceHolderTypes;
+};
+
+// MARK: Unused component
```
bang9, line 9: Should we remove this?

PR #854, opened 2023-11-21
feat: Support custom rendering message menus and reactions

HoonBaek:
[CLNP-1217](https://sendbird.atlassian.net/browse/CLNP-1217)

### ChangeLog
* Add props to the MessageContent component
  * `renderMessageMenu`, `renderEmojiMenu`, and `renderEmojiReactions`

### How to use these?
```tsx
<Channel
  renderMessageContent={(props) => {
    return <MessageContent
      {...props}
      renderMessageMenu={(props) => {
        return <MessageMenu {...props> />
      }}
      renderEmojiMenu={(props) => {
        return <MessageEmojiMenu {...props> />
      }}
      renderEmojiReactions={(props) => {
        return <EmojiReactions {...props> />
      }}
    />
  }}
/>
```

[CLNP-1217]: https://sendbird.atlassian.net/browse/CLNP-1217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/MessageItemMenu/index.tsx
+++ b/src/ui/MessageItemMenu/index.tsx
@@ -21,7 +21,7 @@ import { Role } from '../../lib/types';
 import { ReplyType } from '../../types';
 import { useChannelContext } from '../../modules/Channel/context/ChannelProvider';
 
-interface Props {
+export interface MessageMenuProps {
   className?: string | Array<string>;
   message: SendableMessageType;
   channel: GroupChannel | OpenChannel;
@@ -38,7 +38,7 @@ interface Props {
   onMoveToParentMessage?: () => void;
 }
 
-export default function MessageItemMenu({
+export function MessageMenu({
   className,
   message,
   channel,
@@ -53,7 +53,7 @@ export default function MessageItemMenu({
   setSupposedHover,
   onReplyInThread,
   onMoveToParentMessage = null,
-}: Props): ReactElement {
+}: MessageMenuProps): ReactElement {
   const { stringSet } = useContext(LocalizationContext);
   const triggerRef = useRef(null);
   const containerRef = useRef(null);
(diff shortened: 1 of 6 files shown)


src/ui/MessageItemMenu/index.tsx
```diff
@@ -38,7 +38,7 @@ interface Props {
   onMoveToParentMessage?: () => void;
 }
 
-export default function MessageItemMenu({
+export function MessageMenu({
```
bang9, line 41: Is there a specific reason for changing the component name differently from the file name?

HoonBaek: I tried to match the prop names and the component names
* renderMessageMenu - MessageMenu
* renderEmojiMenu - MessageEmojiMenu

And we were using `export default` so I thought it wouldn't make the backword issue.
But we can make it back if it's not good practice

HoonBaek:
And also pls review the props names

liamcho (approved):
LGTM great job!

src/ui/MessageItemMenu/index.tsx
bang9: I see~ generally, it's always good to ensure that the file name and the component name match, especially for file navigation or stack trace tracking
If you feel that changes are needed, it's a good idea to leave a TODO marker and consider addressing it in v4

AhyoungRyu (commented):
LGTM. 

nit; As I left similar comments, there're several places we can remove duplicated key name - value as of https://eslint.org/docs/latest/rules/object-shorthand.

AhyoungRyu (approved):

PR #856, opened 2023-11-21
feat: Added renderFrozenNotification in ChannelUIProps

liamcho:
Fixes: [CLNP-754](https://sendbird.atlassian.net/browse/CLNP-754)

### Changelogs
- Added `renderFrozenNotification` in `ChannelUIProps`

```
    <Channel
      channelUrl={channelUrl}
      ...
      renderFrozenNotification={() => {
        return (
          <div
            className="sendbird-notification sendbird-notification--frozen sendbird-conversation__messages__notification"
          >My custom Frozen Notification</div>
        );
      }}
    />
```

[CLNP-754]: https://sendbird.atlassian.net/browse/CLNP-754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Channel/components/MessageList/index.tsx
+++ b/src/modules/Channel/components/MessageList/index.tsx
@@ -1,6 +1,6 @@
 import './message-list.scss';
 
-import React from 'react';
+import React, { useMemo } from 'react';
 
 import { useChannelContext } from '../../context/ChannelProvider';
 import PlaceHolder, { PlaceHolderTypes } from '../../../../ui/PlaceHolder';
@@ -28,6 +28,7 @@ export interface MessageListProps {
   renderPlaceholderEmpty?: () => React.ReactElement;
   renderCustomSeparator?: (props: RenderCustomSeparatorProps) => React.ReactElement;
   renderPlaceholderLoader?: () => React.ReactElement;
+  renderFrozenNotification?: () => React.ReactElement;
 }
 
 const MessageList: React.FC<MessageListProps> = ({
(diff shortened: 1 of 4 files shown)


HoonBaek (approved):

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -152,6 +154,13 @@ const MessageList: React.FC<MessageListProps> = ({
 
   const { scrollToBottomHandler, scrollBottom } = useSetScrollToBottom({ loading });
 
+  const renderedFrozenNotification = useMemo(() => {
+    if (renderFrozenNotification) {
+      return renderFrozenNotification();
+    }
+    return null;
+  }, [renderFrozenNotification]);
+
```
bang9, line 163: It seems okay to remove the useMemo.

liamcho: Is it because renderedFrozenNotification does not change often?

bang9: Here are several reasons:

- In typical cases (when `renderFrozenNotification` is not passed), useMemo is unnecessary.
- If `renderFrozenNotification` is passed inline, useMemo has no effect. (always received new function)
- If `renderFrozenNotification` is passed using useCallback, there might be some optimization for component rendering. However, if optimization for component rendering is necessary, it's recommended to use `React.memo` on the side implementing the custom interface rather than using useMemo from our side.

liamcho: Nice reasoning!

bang9 (commented):
lgtm!

PR #859, opened 2023-11-22
fix: Fit the OGMessage Thumbnail Image to the minimal width length

HoonBaek:
Before & After
imageimage

### Fix
* Add width:100% to the ImageRenderer of OGMessage

### ChangeLog
* Fit the OGMessage Thumbnail Image to the minimal width length


liamcho (approved):
nice job! LGTM

PR #870, opened 2023-11-27
fix: Remove the optional chaining of a channel in the GroupChannelList module

HoonBaek:
* Remove the optional chaining of a channel in the GroupChannelList module


bang9 (approved):
lgtm!

PR #874, opened 2023-11-29
fix: Add a message prop to the ReactionItem comp

HoonBaek:
### Fix
Add a message prop to the ReactionItem component for when the Component is not contained by the MessageProvider

### ChangeLog
Add a message prop to the ReactionItem component for when the Component is not contained by the MessageProvider

### DIFF
--- a/src/ui/EmojiReactions/ReactionItem.tsx
+++ b/src/ui/EmojiReactions/ReactionItem.tsx
@@ -41,7 +41,6 @@ export default function ReactionItem({
   const store = useSendbirdStateContext();
   const { isMobile } = useMediaQueryContext();
   const messageStore = useMessageContext();
-  const _message = message ?? messageStore?.message as UserMessage;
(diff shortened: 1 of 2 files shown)


src/ui/EmojiReactions/ReactionItem.tsx
```diff
@@ -49,7 +51,7 @@ export default function ReactionItem({
 
   const handleOnClick = () => {
     setEmojiKey('');
-    toggleReaction?.((message), reaction.key, reactedByMe);
+    toggleReaction?.((_message), reaction.key, reactedByMe);
```
bang9, line 54: How about just passing it inline? Since it's not used in multiple places, there doesn't seem to be a need to create a separate variable for it.

```suggestion
    toggleReaction?.((message ?? messageStore?.message as UserMessage), reaction.key, reactedByMe);
```

bang9 (approved):

PR #878, opened 2023-11-30
performance: Improve image loading speed by using lazy load with IntersectionObserver

liamcho:
Fixes: [<TICKET_ID>](https://sendbird.atlassian.net/browse/<TICKET_ID>)

### Changelogs
- Improved performance of loading image files in `MessageList` with lazy loading technique


src/modules/Channel/components/Message/hooks/useIsElementInViewport.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  useEffect(() => {
+    const observer = new IntersectionObserver(callback);
+    if (elementRef.current) {
+      observer.observe(elementRef.current);
+    }
+    return () => observer.disconnect();
+  }, [elementRef]);
+
+  return { elementRef, isVisible };
```
AhyoungRyu, line 19: Guess we don't need to re-return `elementRef` since there's no places we use it.

src/modules/Channel/components/Message/hooks/useLazyImageLoader.ts
```diff
@@ -0,0 +1,17 @@
...
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  useEffect(() => {
+    if (isLoaded || !isVisible) {
+      return;
+    }
+    setIsLoaded(true);
+  }, [isVisible]);
+
+  return { elementRef, isLoaded };
```
AhyoungRyu, line 16:
```suggestion
  return isLoaded
```
Same for this one.

HoonBaek (commented):
This is a good function!
However, I think we should provide some interface that customers can turn on/off this feature. Because it will change their user experience.
And we need to discuss what the default value will be

src/modules/Channel/components/Message/index.tsx
```diff
@@ -125,6 +126,8 @@ const Message = ({
     setIsScrolled,
   });
 
+  const { isLoaded } = useLazyImageLoader(messageScrollRef);
```
bang9, line 129: It seems that messages unrelated to images are also being re-rendered by the IntersectionObserver.

liamcho:
> This is a good function!
> How about providing an interface that customers can turn on/off this feature? Because it will change their user experience.
> And we need to discuss what the default value will be

I think this is considered a performance issue rather than an option. So I doubt there would be many user who would like to toggle this feature off. So to me it seems unnecesary to add the toggle option.

src/modules/Channel/components/Message/hooks/useIsElementInViewport.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  useEffect(() => {
+    const observer = new IntersectionObserver(callback);
+    if (elementRef.current) {
+      observer.observe(elementRef.current);
+    }
+    return () => observer.disconnect();
+  }, [elementRef]);
+
+  return { isVisible };
```
AhyoungRyu, line 19: If we return an object, it'll trigger unnecessary re-rendering where it's being used. 
Unless there's no specific reason we return an object format in here, we'd better just return `isVisible` instead.

src/modules/Channel/components/Message/hooks/useLazyImageLoader.ts
```diff
@@ -0,0 +1,17 @@
...
+  const { isVisible } = useIsElementInViewport(elementRef);
+
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  useEffect(() => {
+    if (isLoaded || !isVisible) {
+      return;
+    }
+    setIsLoaded(true);
+  }, [isVisible]);
```
AhyoungRyu, line 14: Same here :)

src/modules/Channel/components/Message/hooks/useLazyImageLoader.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  return { isLoaded };
```
AhyoungRyu, line 16: Same here :)

src/modules/Channel/components/Message/hooks/useIsElementInViewport.ts
liamcho: ah i miss this thanks!

bang9 (approved):

bang9 (approved):
lgtm!

PR #880, opened 2023-12-01
fix: Divide the EditUserProfileUI to Modal and View parts

HoonBaek:
[CLNP-1656](https://sendbird.atlassian.net/browse/CLNP-1656)

## Issue
A customer tried to use the props `renderUserProfile` like below
```tsx
 <Channel
   renderUserProfile={() => (
      <EditUserProfileProvider ...>
        <EditUserProfileUI />
      </EditUserProfileProvider>
   )}
 />
```
and faced an issue that the EditUserProfile Modal is closed even when clicking the Modal contents.

This issue has occurred because [multiple Portals](https://react.dev/reference/react-dom/createPortal) are used inside of [our implementation](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/ui/MessageContent/MessageProfile/index.tsx#L62)[(MenuItems)](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/ui/ContextMenu/MenuItems.tsx) and the [EditUserProfileUI](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/modules/EditUserProfile/components/EditUserProfileUI/index.tsx) component twice.

## Solution
We can resolve this issue, so we are providing a presentation layer component, the `EditUserProfileUIView` component that customers can use instead of the `EditUserProfileUI` component.
#### Use EditUserProfile
If you want to show the EditUserProfile on our Modal component, use like this
```tsx
import EditUserProfile from '@sendbird/uikit-react/EditUserProfile'

 <Channel
   renderUserProfile={() => (
      <EditUserProfile ... />
   )}
 />
```

#### Use EditUserProfileUIView
If you want to implement the EditUserProfile page on your own place, use the EditUserProfileUIView component instead
```tsx
import { EditUserProfileUIView, useEditUserProfileUISates } from '@sendbird/uikit-react/EditUserProfile/components/EditUserProfileUI';

 <Channel
   renderUserProfile={(props) => (
     <CustomUserProfile {...props}/>
   )}
 />
 
 const CustomUserProfile = ({
   close,
   user,
   currentUserId,
 }) => {   
   const onEditProfile = (updatedUser: User) => { /* custom logic */ };
   const onThemeChange = (them: string) => { /* custom logic */ };
   
   const {
     formRef,
     inputRef,
     updateUserInfo,
     setProfileImage,
   } = useEditUserProfileUISates({ onEditProfile });
   
   return (
       <>{/* custom wrapper */}
         <EditUserProfileUIView
           formRef={formRef}
           inputRef={inputRef}
           setProfileImage={setProfileImage}
           onThemeChange={onThemeChange}
         />
         <button
           value="cancel"
           onClick={close}
         />
         <button
           value="submit"
           onClick={updateUserInfo}
         />
       </>
   )
 }
```

## Fix&ChangeLog
* Take out the `renderUserProfile` from the MenuItems component in the MessageProfile comp
* Divide the `EditUserProfileUI` component to
  * Modal & Logic part
  * A presentation layer: EditUserProfileUIView component
* Removed the default value of 'left' props from the ui/Avatar component

[CLNP-1656]: https://sendbird.atlassian.net/browse/CLNP-1656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/EditUserProfile/components/EditUserProfileUI/EditUserProfileUIView.tsx
+++ b/src/modules/EditUserProfile/components/EditUserProfileUI/EditUserProfileUIView.tsx
@@ -1,0 +1,135 @@
+import React, { useState, useRef, type MutableRefObject, type Dispatch } from 'react'
+
+import Input, { InputLabel } from '../../../../ui/Input';
+import { useLocalization } from '../../../../lib/LocalizationContext';
+import Avatar from '../../../../ui/Avatar';
+import TextButton from '../../../../ui/TextButton';
+import Label, { LabelColors, LabelTypography } from '../../../../ui/Label';
+import Icon, { IconTypes } from '../../../../ui/Icon';
+import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
+
+export interface EditUserProfileUIViewProps {
+  formRef: MutableRefObject<any>;
+  inputRef: MutableRefObject<any>;
+  onThemeChange: (theme: string) => void;
+  setProfileImage: Dispatch<File | null>;
+}
+export const EditUserProfileUIView = ({
+  formRef,
+  inputRef,
+  onThemeChange,
+  setProfileImage,
+}: EditUserProfileUIViewProps) => {
+  const { stores, config } = useSendbirdStateContext();
+  const { theme, setCurrentTheme } = config;
+  const user = stores.userStore?.user;
+  const { stringSet } = useLocalization();
+
+  const [currentImg, setCurrentImg] = useState(null);
+  const hiddenInputRef = useRef(null);
+
+  return (
+    <form
+      className="sendbird-edit-user-profile"
+      ref={formRef}
+      onSubmit={(e) => { e.preventDefault(); }}
+    >
+      <section className="sendbird-edit-user-profile__img">
+        <InputLabel>
+          {stringSet.EDIT_PROFILE__IMAGE_LABEL}
+        </InputLabel>
+        <div className="sendbird-edit-user-profile__img__avatar">
+          <Avatar
+            width="80px"
+            height="80px"
+            src={currentImg || user?.profileUrl}
+          />
+        </div>
+        <input
+          ref={hiddenInputRef}
+          type="file"
+          accept="image/gif, image/jpeg, image/png"
+          style={{ display: 'none' }}
+          onChange={(e) => {
+            setCurrentImg(URL.createObjectURL(e.target.files[0]));
+            setProfileImage(e.target.files[0]);
+            hiddenInputRef.current.value = '';
+          }}
+        />
+        <TextButton
+          className="sendbird-edit-user-profile__img__avatar-button"
+          disableUnderline
+          onClick={() => hiddenInputRef.current.click()}
+        >
+          <Label
+            type={LabelTypography.BUTTON_1}
+            color={LabelColors.PRIMARY}
+          >
+            {stringSet.EDIT_PROFILE__IMAGE_UPLOAD}
+          </Label>
+        </TextButton>
+      </section>
+      <section className="sendbird-edit-user-profile__name">
+        <InputLabel>
+          {stringSet.EDIT_PROFILE__NICKNAME_LABEL}
+        </InputLabel>
+        <Input
+          required={user?.nickname !== ''}
+          name="sendbird-edit-user-profile__name__input"
+          ref={inputRef}
+          value={user?.nickname}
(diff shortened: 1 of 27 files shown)


HoonBaek:
The PR became too big, let me separate and create one more PR for the changes which are related to removing the default value 'left' from Avatar

src/modules/EditUserProfile/components/EditUserProfileUI/EditUserProfileUIView.tsx
```diff
@@ -0,0 +1,135 @@
...
+          {
+            theme === 'dark'
+              ? (
+                <Icon
+                  onClick={() => {
+                    setCurrentTheme('light');
+                    onThemeChange?.('light');
+                    // if (onThemeChange && typeof onThemeChange === 'function') {
+                    //   onThemeChange('light');
+                    // }
```
liamcho, line 109: If we are going to keep comments, we should add explanation too. Otherwise lets remove it!

src/modules/EditUserProfile/components/EditUserProfileUI/EditUserProfileUIView.tsx
```diff
@@ -0,0 +1,135 @@
...
+                />
+              )
+              : (
+                <Icon
+                  onClick={() => {
+                    setCurrentTheme('dark');
+                    onThemeChange?.('dark');
+                    // if (onThemeChange && typeof onThemeChange === 'function') {
+                    //   onThemeChange('dark');
+                    // }
```
liamcho, line 123: If we are going to keep comments, we should add explanation too. Otherwise lets remove it!

src/modules/EditUserProfile/components/EditUserProfileUI/EditUserProfileUIView.tsx
HoonBaek: Oh, yes thanks

src/modules/EditUserProfile/components/EditUserProfileUI/index.tsx
```diff
@@ -1,47 +1,100 @@
...
+  inputRef,
+  profileImage,
+  onEditProfile,
+}: HandleUpdateUserInfoParams) => {
+  const { stores, dispatchers } = globalContext;
+  const sdk = stores.sdkStore.sdk;
+  const user = stores.userStore.user;
+  const { userDispatcher } = dispatchers;
+
+  if (user?.nickname !== '' && !inputRef.current.value) {
```
liamcho, line 40: why we need to check empty nickname here?

src/modules/OpenChannelSettings/components/ParticipantUI/ParticipantItem.tsx
```diff
@@ -65,32 +65,31 @@ export const UserListItem: React.FC<UserListItemProps> = ({
...
+              })
+              : (
+                <MenuItems
+                  openLeft
+                  parentRef={avatarRef}
+                  // for catching location(x, y) of MenuItems
+                  parentContainRef={avatarRef}
+                  // for toggling more options(menus & reactions)
+                  closeDropdown={closeDropdown}
+                  style={{ paddingTop: '0px', paddingBottom: '0px' }}
```
liamcho, line 83: I see repeating
```
{ paddingTop: '0px', paddingBottom: '0px' }
```
can we make this default? it not why not?

liamcho (changes requested):

src/modules/EditUserProfile/components/EditUserProfileUI/index.tsx
HoonBaek: Because it's meanless to update the user nickname with empty string, so we return nothing when it's empty

liamcho: so its || not && ?

HoonBaek: The `user.nickname` is for the nickname
and the `inputRef.current.value` is for the profile image

So we can ignore updating when both inputs are not given

src/modules/OpenChannelSettings/components/ParticipantUI/ParticipantItem.tsx
HoonBaek: That's a good suggestion, but I'm little bit worried about the side effects, let me do it in a separate PR

liamcho (approved):
LG2M

PR #881, opened 2023-12-05
bug-fix: Group channel user left or banned event should not be ignored

liamcho:
Fixes: [<TICKET_ID>](https://sendbird.atlassian.net/browse/<TICKET_ID>)

### Changelogs
- Fixed a bug where  `currentGroupChannel` of `useChannelContext()` not being updated when non-current user has left or been banned while `ChannelList` is not being used


HoonBaek (approved):
Good catch!

PR #888, opened 2023-12-11
fix: Use PropsWithChildren<unknown> instead PropsWithChildren

HoonBaek:
### Issue
* [Can't use `PropsWithChildren` because of type error](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60641)
  * Not assignable to type error, property 'children' doesn't exist on type `PropsWithChildren`
* [Can't use `PropsWithChildren<{}>` because of lint error](https://stackoverflow.com/questions/65548388/react-propswithchildren-with-no-props-other-than-children)

### Fix
* Use Use `PropsWithChildren<unknown>` instead `PropsWithChildren`

### ChangeLog
* Fix a type error from SendbirdProvider
  * Not assignable to type error, property 'children' doesn't exist on type `PropsWithChildren`

### DIFF
--- a/src/hooks/useModal/index.tsx
+++ b/src/hooks/useModal/index.tsx
@@ -8,7 +8,7 @@ export type OpenGlobalModalProps = {
   modalProps: ModalProps;
   childElement: (props: { closeModal: () => void }) => ReactElement;
 };
-export type GlobalModalProviderProps = React.PropsWithChildren;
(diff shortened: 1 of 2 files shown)


bang9 (approved):
lgtm

HoonBaek:
Hi @ashr81 ! Thanks for your interest in our project, we are going to release the `v3.9.2` tmr. Thank you!

PR #894, opened 2023-12-13
fix: channel view should show NO_CHANNEL placeholder after leaving all channels

aldenquimby:
- before change - after leaving your only channel, you still see a channel view with `No Title` that allows typing messages (but then fails to send):
before

- after change - you see the NO_CHANNEL placeholder, as expected:
after

# CLA Notes
- as noted on #893 , I've reached out to support asking how to move forward with a CLA


AhyoungRyu:
Thank you for your contribution @aldenquimby the change looks good to me.
But unfortunately we're not accepting external code change due to our policy. Let me create a new PR instead. Thanks again!

PR #895, opened 2023-12-14
fix: scroll issues

bang9:
- fix: maintain scroll position when loading previous messages
- fix: maintain scroll position when loading next messages
- fix: move the logic that delay for rendering mmf to the correct location

ticket: [CLNP-1735]

### DIFF
--- a/src/modules/Channel/components/Message/index.tsx
+++ b/src/modules/Channel/components/Message/index.tsx
@@ -28,7 +28,6 @@ import RemoveMessageModal from '../RemoveMessageModal';
 import { MessageInputKeys } from '../../../../ui/MessageInput/const';
 import { EveryMessage, RenderCustomSeparatorProps, RenderMessageProps } from '../../../../types';
 import { useLocalization } from '../../../../lib/LocalizationContext';
-import { useHandleOnScrollCallback } from '../../../../hooks/useHandleOnScrollCallback';
 import { useDirtyGetMentions } from '../../../Message/hooks/useDirtyGetMentions';
 import SuggestedReplies from '../SuggestedReplies';
 
@@ -93,8 +92,6 @@ const Message = ({
     onQuoteMessageClick,
     onMessageAnimated,
     onMessageHighlighted,
-    onScrollCallback,
-    setIsScrolled,
     sendMessage,
     localMessages,
   } = useChannelContext();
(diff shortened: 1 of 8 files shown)


src/modules/Channel/components/Message/index.tsx
```diff
@@ -162,7 +152,7 @@ const Message = ({
     let animationTimeout = null;
     let messageHighlightedTimeout = null;
     if (highLightedMessageId === message.messageId && messageScrollRef?.current) {
-      handleOnScroll();
+      messageScrollRef.current.scrollIntoView({ block: 'center', inline: 'center' });
```
bang9, line 155: revert https://github.com/sendbird/sendbird-uikit-react/pull/492

src/modules/Channel/components/MessageList/index.tsx
```diff
@@ -89,27 +91,11 @@ const MessageList: React.FC<MessageListProps> = ({
...
-              1000,
-            );
-            // element.scrollTop = scrollHeight - clientHeight;
-            // scrollRef.current.scrollTop = scrollHeight - clientHeight;
-          } catch (error) {
-            //
-          }
-        }
-      });
+      onScrollDownCallback(callback);
```
bang9, line 98: fix https://github.com/sendbird/sendbird-uikit-react/pull/810

HoonBaek (approved):
Really appreciate for this job!

src/modules/Channel/components/MessageList/hooks/useSetScrollToBottom.ts
```diff
@@ -20,13 +16,15 @@ export function useSetScrollToBottom({
...
+        element.scrollHeight - element.scrollTop - element.offsetHeight,
+      );
     } catch {
       //
     }
   };
   return {
     scrollBottom,
-    scrollToBottomHandler: useDebounce(scrollCb, DELAY),
+    scrollToBottomHandler: useThrottleCallback(scrollCb, DELAY, { trailing: true }),
```
bang9, line 28: This has caused a slow update of the scrollBottom value.
The scrollBottom value is used in a place where adjustments are made when the size of the content within the current scroll view changes, and as a result, it was not being updated with the latest values.
(e.g., When replying, the scroll does not go to the bottom.)

AhyoungRyu (approved):

PR #898, opened 2023-12-20
feat: Feedback message

liamcho:
Fixes: [AC-745](https://sendbird.atlassian.net/browse/AC-745)

### Feedback Message
Now we are supporting Feedback Message feature!<br/>
Feedback feature is applied to message with non default `myFeedbackStatus` value.<br/>

* Added `MessageFeedbackModal`, `FeedbackIconButton`, and `MobileFeedbackMenu`

### Documentation
- https://sendbird.atlassian.net/wiki/spaces/SDK/pages/2184347696/Feedback+for+message

### Figma design
- https://www.figma.com/file/SVbXU00FhjztekD8AiVukK/UIKit_Work-file_React?node-id=1819%3A12270&mode=dev

### GIFs
Will be uploaded soon.

[AC-745]: https://sendbird.atlassian.net/browse/AC-745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
```diff
@@ -0,0 +1,46 @@
+import { useCallback, useLayoutEffect } from 'react';
+
+type KeyDownCallbackMap = Record<string, (event: React.KeyboardEvent<HTMLDivElement>) => void>;
+
+interface UseKeyDownV2Props {
+  ref: React.RefObject<HTMLDivElement>;
+  keyDownCallbackMap: KeyDownCallbackMap;
+}
+
+export function useKeyDownV2(props: UseKeyDownV2Props): React.KeyboardEventHandler<HTMLDivElement> {
```
AhyoungRyu, line 10: Q. What's the reason it has V2 suffix?

src/modules/Channel/components/MessageFeedbackModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        customFooter={
+          <div className='sendbird-message-feedback-modal-footer__root'>
+            {
+              !isMobile && message?.myFeedback && selectedFeedback === message.myFeedback.rating
+                ? <Button type={ButtonTypes.WARNING} onClick={onRemove}>
+                  <Label type={LabelTypography.BUTTON_3} color={LabelColors.ERROR}>
+                    {stringSet.BUTTON__REMOVE_FEEDBACK}
+                  </Label>
+                </Button>
+                : <div/>
```
AhyoungRyu, line 73: It looks a bit awkward that a closing tag is next to the `:` ..?

src/styles/_color-themes.scss
```diff
@@ -157,6 +157,7 @@ $themes: (
     primary--1-bg--7: var(--sendbird-dark-background-700),
     secondary--3-2: var(--sendbird-dark-secondary-200),
     error--3-2: var(--sendbird-dark-error-200),
+
```
AhyoungRyu, line 160: Let's remove this change if it's not intended :)

src/ui/MessageContent/index.tsx
```diff
@@ -149,6 +158,11 @@ export default function MessageContent(props: MessageContentProps): ReactElement
   // Thread replies
   const displayThreadReplies = message?.threadInfo?.replyCount > 0 && replyType === 'THREAD';
 
+  // Feedback buttons
+  // TODO: Replace line below with comment before merging.
+  const isFeedbackMessage = !isByMe; // !isByMe && message?.myFeedbackStatus && message.myFeedbackStatus !== FeedbackStatus.NOT_APPLICABLE;
+  const isFeedbackMessageClassName = isFeedbackMessage ? 'sendbird-message-content__feedback' : '';
```
AhyoungRyu, line 164:
```suggestion
  const feedbackMessageClassName = isFeedbackMessage ? 'sendbird-message-content__feedback' : '';
```

Probably you meant this?

src/ui/MessageContent/index.tsx
```diff
@@ -390,6 +445,63 @@ export default function MessageContent(props: MessageContentProps): ReactElement
...
+              if (selectedFeedback) {
+                if (message.myFeedback) {
+                  if (comment !== message.myFeedback.comment) {
+                    const newFeedback: Feedback = new Feedback({
+                      id: message.myFeedback.id,
+                      rating: selectedFeedback,
+                      comment,
+                    });
+                    await message.updateFeedback(newFeedback);
+                  }
```
AhyoungRyu, line 482:
```suggestion
                if (!message.myFeedback) {
                  return;
                }
                if (comment !== message.myFeedback.comment) {
                  const newFeedback: Feedback = new Feedback({
                    id: message.myFeedback.id,
                    rating: selectedFeedback,
                    comment,
                  });
                  await message.updateFeedback(newFeedback);
                }
```

There're 3 nested if stmts here. How about we use the early-return pattern like the above for the better reabability? It'll reduce 1 level ;)

src/ui/MobileFeedbackMenu/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+        <div
+          className='sendbird-message__bottomsheet--feedback-option'
+          onClick={() => {
+            hideMenu();
+            onRemoveFeedback();
+          }}
+        >
+          <Label type={LabelTypography.BODY_1} color={LabelColors.ERROR}>
+            {stringSet?.REMOVE_FEEDBACK}
```
AhyoungRyu, line 46:
```suggestion
            {stringSet.EDIT_COMMENT}
          </Label>
        </div>
        <div
          className='sendbird-message__bottomsheet--feedback-option'
          onClick={() => {
            hideMenu();
            onRemoveFeedback();
          }}
        >
          <Label type={LabelTypography.BODY_1} color={LabelColors.ERROR}>
            {stringSet.REMOVE_FEEDBACK}
```
I guess we don't need to use the optional chaining here

AhyoungRyu (commented):
Just left minor comments.
Great work. Thanks @liamcho! 🎉

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: useKeyDown.ts already exists but the old one is not compatible and this new one is the better version.

src/modules/Channel/components/MessageFeedbackModal/index.tsx
liamcho: Ah its an empty div: <div></div>. I tried with <></> but its not an element so breaks the intended flex layout. So I had to use empty div instead.

src/ui/MessageContent/index.tsx
liamcho: oh no, if `!message.myFeedback`, we still need to do two actions:
1. submitFeedback.
2. setSelectedFeedback(null);

And setSelectedFeedback is a common action so I think its better to keep it as is. What do you think?

AhyoungRyu: Oh there's another logic below the block. Still I think we can do something. Then how about this?
```
if (!selectedFeedback) { // let's separate the first if block
  return;
}
if (message.myFeedback) {
  ...
} else {
  ...
}
setShowFeedbackModal(false);
```

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
AhyoungRyu: Yeah I saw we have useKeyDown already. But it seems they are not being replaceable but being used in separate places.

src/modules/Channel/components/MessageFeedbackModal/index.tsx
AhyoungRyu: Then 
```
!isMobile && message?.myFeedback && selectedFeedback === message.myFeedback.rating
                ? <Button type={ButtonTypes.WARNING} onClick={onRemove}>
                  <Label type={LabelTypography.BUTTON_3} color={LabelColors.ERROR}>
                    {stringSet.BUTTON__REMOVE_FEEDBACK}
                  </Label>
                </Button>
```
^ this block won't have a closing div tag.

How about this? Even it has a little duplicated class names, but it looks more natural I think.  
```
customFooter={
// To display the delete feedback button on the wide screen
!isMobile && message?.myFeedback && selectedFeedback === message.myFeedback.rating
  ? <div className='sendbird-message-feedback-modal-footer__root'>...</div>
  : <div className='sendbird-message-feedback-modal-footer__root' />
```

src/ui/MessageContent/index.tsx
liamcho: Yes that is better 👍

src/modules/Channel/components/MessageFeedbackModal/index.tsx
liamcho: if we were to take out 
```
!isMobile && message?.myFeedback && selectedFeedback === message.myFeedback.rating
```
outside of
```
<div className='sendbird-message-feedback-modal-footer__root'>
```
then we need to have
```
<div className='sendbird-message-feedback-modal-footer__right-content'>
```
in both web and mobile `<div>` which creates duplicated code. Are we ok with this or did I miss something?

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: Yes their goals are the same but V2 is a better one. From now on we should use V2 instead. We can remove V1 later.

src/modules/Channel/components/MessageFeedbackModal/index.tsx
AhyoungRyu: Oh sorry for the delay. Yeah I saw the full code and now I got your point. Hmm it's weird that we need `<div />` not to break the layout. We probably need to see the style setting.

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
bang9, line 10: How about simplifying the parameter interface of the hook for easier usage?

```ts
function useKeydown(ref: React.RefObject<HTMLDivElement>, listener: KeyDownCallbackMap) { ... }
```

```ts
// AS-IS
useKeydown({
  ref: elemRef,
  keyDownCallbackMap: {
    Escape: () => { ... },
    ArrowLeft: () => { ... },
    ArrowRight: () => { ... },
    Enter: () => { ... },
  }
})

// TO-BE
useKeydown(elemRef, {
  Escape: () => { ... },
  ArrowLeft: () => { ... },
  ArrowRight: () => { ... },
  Enter: () => { ... },
})
```

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
```diff
@@ -0,0 +1,46 @@
...
+      case 'Enter': {
+        const callback = keyDownCallbackMap['Enter'];
+        callback?.(event);
+        break;
+      }
+      default:
+        break;
+    }
+    event.stopPropagation();
+  }, []);
```
bang9, line 43: How about just running it without switch-case?
```ts
const callback = keyDownCallbackMap[event.key];
callback?.(event);
```

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function useKeyDownV2(props: UseKeyDownV2Props): React.KeyboardEventHandler<HTMLDivElement> {
+  const { ref, keyDownCallbackMap } = props;
+
+  useLayoutEffect(() => {
+    ref.current?.focus();
+  }, [ref.current]);
+
+  const onKeyDown: React.KeyboardEventHandler<HTMLDivElement> = useCallback((event) => {
```
bang9, line 17:
```suggestion
  const onKeyDown: React.KeyboardEventHandler<HTMLDivElement> = usePreservedCallback((event) => {
```

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: @AhyoungRyu I refactored `useKeyDown.ts` !

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: oh nice!

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: SG2M!

liamcho: ah actually we are trying to benefit by using
```
const callback = keyDownCallbackMap[event.key];
callback?.(event);
```
so lets keep as is

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
liamcho: @bang9 keyDownCallbackMap must be a dependency so I think we have to use useCallback here

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
bang9: Oh, the object form will be retained.

```ts
// AS-IS
const callback = params.keyDownCallbackMap[event.key];
callback?.(event);

// TO-BE
const callback = keyDownCallbackMap[event.key]; // callbacks[event.key]
callback?.(event);
```

liamcho: did you mean to prefer
```
const callback = props[event.key];
```
?

liamcho: Also we want to use `keyDownCallbackMap` as dependency for `useCallback` so I think we had better keep as is
```
interface UseKeyDownProps {
  ref: React.RefObject<HTMLDivElement>;
  keyDownCallbackMap: KeyDownCallbackMap;
}
```

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
bang9: This is the issue. using `usePreservedCallback` instead of `useCallback` will resolve this issue.

```tsx
useKeydown({
  ref,
  keyDownCallbackMap: {    // --> Inline object ==> Always new object
    Escape: () => { ... }
  }
})

const useKeydown = ({ keyDownCallbackMap }) => {
  const onKeyDown = useCallback = useCallback(() => {
    // ...
  },[keyDownCallbackMap]) // --> Always new object
  
  return onKeyDown // --> Always new callback
}
```

liamcho: oh it seems to work. That's very interesting that usePreservedCallback does not need dependency but useCallback does

src/modules/Channel/components/MessageFeedbackModal/hooks/useKeyDownV2.ts
bang9: Ah, the reason I suggested it is because I felt it unnecessary to specify the `keyDownCallbackMap` key when using the hook.

```tsx
// AS-IS
useKeydown({
  ref: elemRef,
  keyDownCallbackMap: { // -> here
    Escape: () => { ... },
    ArrowLeft: () => { ... },
    ArrowRight: () => { ... },
    Enter: () => { ... },
  }
})

// TO-BE
useKeydown(elemRef, {
  Escape: () => { ... },
  ArrowLeft: () => { ... },
  ArrowRight: () => { ... },
  Enter: () => { ... },
})
```

Regarding this, there are no suggestions for changes in the internal logic.

liamcho: gotcha!

src/ui/FileViewer/__tests__/useKeyDown.spec.tsx
bang9, line 1: This test file seems to be no longer necessary.

src/ui/MessageContent/index.tsx
```diff
@@ -149,6 +158,11 @@ export default function MessageContent(props: MessageContentProps): ReactElement
   // Thread replies
   const displayThreadReplies = message?.threadInfo?.replyCount > 0 && replyType === 'THREAD';
 
+  // Feedback buttons
+  // TODO: Replace line below with comment before merging.
+  const isFeedbackMessage = !isByMe; // !isByMe && message?.myFeedbackStatus && message.myFeedbackStatus !== FeedbackStatus.NOT_APPLICABLE;
```
bang9, line 163: Do we need to apply this todo?

src/ui/MessageContent/index.tsx
```diff
@@ -38,6 +38,10 @@ import { noop } from '../../utils/utils';
 import MessageProfile, { MessageProfileProps } from './MessageProfile';
 import MessageBody, { MessageBodyProps } from './MessageBody';
 import MessageHeader, { MessageHeaderProps } from './MessageHeader';
+import Icon, { IconTypes } from '../Icon';
+import FeedbackIconButton from '../FeedbackButton';
```
bang9, line 42:
```suggestion
import FeedbackButton from '../FeedbackButton';
```

src/ui/MessageContent/index.tsx
```diff
@@ -301,6 +315,47 @@ export default function MessageContent(props: MessageContentProps): ReactElement
               }
             </div>
           )}
+          {/* Feedback buttons */}
+          {
+            isFeedbackMessage && <div
+              className='sendbird-message-content__middle__body-container__feedback-buttons-container'
+            >
+              <FeedbackIconButton
```
bang9, line 323:
```suggestion
              <FeedbackButton
```

src/ui/MessageContent/index.tsx
```diff
@@ -301,6 +315,47 @@ export default function MessageContent(props: MessageContentProps): ReactElement
...
+                  }
+                }}
+              >
+                <Icon
+                  type={IconTypes.FEEDBACK_LIKE}
+                  width='24px'
+                  height='24px'
+                />
+              </FeedbackIconButton>
+              <FeedbackIconButton
```
bang9, line 340:
```suggestion
              </FeedbackButton>
              <FeedbackButton
```

src/ui/MessageContent/index.tsx
```diff
@@ -301,6 +315,47 @@ export default function MessageContent(props: MessageContentProps): ReactElement
...
+                    setShowFeedbackModal(true);
+                  }
+                }}
+              >
+                <Icon
+                  type={IconTypes.FEEDBACK_DISLIKE}
+                  width='24px'
+                  height='24px'
+                />
+              </FeedbackIconButton>
```
bang9, line 356:
```suggestion
              </FeedbackButton>
```

bang9 (approved):
lgtm

src/ui/MessageContent/index.tsx
liamcho: Thanks for the reminder ill put this back

src/ui/MessageContent/index.tsx
liamcho: oh thanks but its FeedbackIconButton. I changed to below instead.
```
import FeedbackIconButton from '../FeedbackIconButton';

```

src/ui/MessageContent/index.tsx
liamcho: above

src/ui/MessageContent/index.tsx
liamcho: above

src/ui/MessageContent/index.tsx
liamcho: above

AhyoungRyu (approved):

bang9 (approved):

PR #899, opened 2023-12-20
refactor: move --sendbird-vh css variable setting logic to comp

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-1806

We have been calculating the height value of the modules/InviteUsers modal using the --sendbird-vh CSS variable in the following manner: [GitHub Link](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/modules/CreateChannel/components/InviteUsers/invite-users.scss#L13).

I believe this approach was taken to adjust the dynamic scrollable area height, especially in mobile view. However, it hampers code readability as we need to search for the keyword to locate where it is defined.

Therefore, I propose relocating this logic to the InviteUsers component and consolidating all related logic within a single file.

Confirmed that it's working as before on mobile window size change. 
record

### DIFF
--- a/src/modules/CreateChannel/components/InviteUsers/index.tsx
+++ b/src/modules/CreateChannel/components/InviteUsers/index.tsx
@@ -54,7 +54,7 @@ const InviteUsers: React.FC<InviteUsersProps> = ({
   const titleText = stringSet.MODAL__CREATE_CHANNEL__TITLE;
   const submitText = stringSet.BUTTON__CREATE;
   const { isMobile } = useMediaQueryContext();
-  const [appHeight, setAppHeight] = useState<number>(window.innerHeight);
+  const [scrollableAreaHeight, setScrollableAreaHeight] = useState<number>(window.innerHeight);
 
   const userQueryCreator = userListQuery ? userListQuery() : createDefaultUserListQuery({ sdk });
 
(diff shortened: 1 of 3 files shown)


src/modules/CreateChannel/components/InviteUsers/index.tsx
```diff
@@ -76,11 +69,11 @@ const InviteUsers: React.FC<InviteUsersProps> = ({
     }
   }, []);
 
-  // https://stackoverflow.com/a/70302463
-  // https://css-tricks.com/the-trick-to-viewport-units-on-mobile/#css-custom-properties-the-trick-to-correct-sizing
-  // to fix navbar break in mobile
+  // To fix navbar break in mobile we set dynamic height to the scrollable area
   useEffect(() => {
-    appHeight();
+    const appHeight = () => {
```
liamcho, line 74: Just minore: let's change this name to something other than `appHeight` which is a state.

liamcho (approved):
LG2M

src/modules/CreateChannel/components/InviteUsers/index.tsx
AhyoungRyu: Yeah sounds good https://github.com/sendbird/sendbird-uikit-react/pull/899/commits/3a8b7ea7f9802ba5c7091294d89c4cbc0fe971b1

PR #900, opened 2023-12-21
fix: prevent access to window in SSR environment

AhyoungRyu:
Fixes https://sendbird.atlassian.net/browse/SBISSUE-14287 

The change is originally made by @AaronJamesKing 👍 https://github.com/notice-co/sendbird-uikit-react/commit/e393069793b344cf78c7cbc9db48b1473aeac7a4

### DIFF
--- a/src/lib/hooks/useOnlineStatus.ts
+++ b/src/lib/hooks/useOnlineStatus.ts
@@ -6,7 +6,12 @@ import { uuidv4 } from '../../utils/uuid';
 import { LoggerInterface } from '../Logger';
 
 function useOnlineStatus(sdk: SendbirdChat, logger: LoggerInterface) {
-  const [isOnline, setIsOnline] = useState(window?.navigator?.onLine ?? true);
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useOnlineStatus.ts
```diff
@@ -6,7 +6,12 @@ import { uuidv4 } from '../../utils/uuid';
 import { LoggerInterface } from '../Logger';
 
 function useOnlineStatus(sdk: SendbirdChat, logger: LoggerInterface) {
-  const [isOnline, setIsOnline] = useState(window?.navigator?.onLine ?? true);
+  const [isOnline, setIsOnline] = useState(
+    // window is undefined in SSR env
+    window != null
```
bang9, line 11: It seems like an error may occur because the "window" is not defined.
We might need to use `typeof window !== 'undefined'` to prevent this error.

liamcho (approved):
LGTM

bang9 (approved):

PR #901, opened 2023-12-21
fix: channel view should show NO_CHANNEL placeholder after leaving all channels

AhyoungRyu:
The change comes from #894 and made by @aldenquimby

### DIFF
--- a/src/modules/ChannelList/context/ChannelListProvider.tsx
+++ b/src/modules/ChannelList/context/ChannelListProvider.tsx
@@ -311,7 +311,13 @@ const ChannelListProvider: React.FC<ChannelListProviderProps> = (props: ChannelL
 
   // Set current channel (by on_channel_selected event)
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


aldenquimby:
Thank you @AhyoungRyu !

bang9 (approved):

aldenquimby:
@AhyoungRyu @bang9 FYI I believe this caused an issue with the mobile layout in the default App:
https://github.com/sendbird/sendbird-uikit-react/blob/0f8fc3a639fe3f82fc44873d684442b100328cb1/src/modules/App/MobileLayout.tsx#L109-L112

I think this needs to change to:
```ts
            onChannelSelect={(channel) => {
              setCurrentChannel(channel);
              if (channel) {
                setPanel(PANELS.ChannelPanel);
              }
            }}
```

So that you only switch to the channel panel if a channel was actually selected (vs unselected / left)

Apologies for not catching this in my initial PR

aldenquimby:
@AhyoungRyu @bang9 another possibly related bug here (although it looks to me like it may have been a problem beforehand too): #937

PR #903, opened 2024-01-02
fix: Do not upsize the images in ImageCompression

HoonBaek:
### Issue
[CLNP-1832](https://sendbird.atlassian.net/browse/CLNP-1832)
* Image Compression feature upsizes the images when the resizing values are greater than the origin length

### Fix
* Add optional chaining for the case when a logger doesn't exist
* Do not use the resizing value if it's greater than the origin length

### ChangeLog
* Do not upsize the images when the resizing values are greater than the origin length

[CLNP-1832]: https://sendbird.atlassian.net/browse/CLNP-1832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/compressImages.ts
+++ b/src/utils/compressImages.ts
@@ -9,7 +9,7 @@ interface CompressImageParams {
   resizingHeight?: number;
 }
 
-const compressImage = ({
+export const compressImage = ({
   imageFile,
   compressionRate,
   resizingWidth,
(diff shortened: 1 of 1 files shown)


src/utils/compressImages.ts
```diff
@@ -25,13 +25,15 @@ const compressImage = ({
 
       const originWidth = image.width;
       const originHeight = image.height;
-      const widthRatio = originWidth / (resizingWidth || originWidth);
-      const heightRatio = originHeight / (resizingHeight || originHeight);
+      let targetResizingWidth = (!resizingWidth || resizingWidth > originWidth) ? originWidth : resizingWidth;
+      let targetResizingHeight = (!resizingHeight || resizingHeight > originHeight) ? originHeight : resizingHeight;
```
AhyoungRyu, line 29:
```suggestion
      const targetResizingWidth = (!resizingWidth || resizingWidth > originWidth) ? originWidth : resizingWidth;
      const targetResizingHeight = (!resizingHeight || resizingHeight > originHeight) ? originHeight : resizingHeight;
```
I think it's okay to keep them as const since they won't be re-assigned.

AhyoungRyu (approved):

src/utils/compressImages.ts
HoonBaek: They have a chance of being re-assigned on the line 38 and 40 😁

AhyoungRyu: Ahh right. I missed that.

PR #907, opened 2024-01-05
bug-fix: Fix cross site scripting vulnerability in OGTag

liamcho:
Fixes: [CLNP-1845](https://sendbird.atlassian.net/browse/CLNP-1845)

### Changelogs
- Fixed cross site scripting vulnerability in OGTag

[CLNP-1845]: https://sendbird.atlassian.net/browse/CLNP-1845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/OGMessageItemBody/index.tsx
+++ b/src/ui/OGMessageItemBody/index.tsx
@@ -37,8 +37,15 @@ export default function OGMessageItemBody({
 }: Props): ReactElement {
   const imageRef = useRef<HTMLDivElement>(null);
   const { stringSet } = useContext(LocalizationContext);
+
   const openOGUrl = (): void => {
-    if (message?.ogMetaData?.url) window.open(message?.ogMetaData?.url);
+    let url = message?.ogMetaData?.url;
+    if (url) {
+      if (!url.startsWith('http')) {
+        url = 'https://' + url;
+      }
+      window.open(url);
+    }
   };
   const isMessageMentioned = isMentionEnabled && message?.mentionedMessageTemplate?.length > 0 && message?.mentionedUsers?.length > 0;
   const tokens = useMemo(() => {
--- a/src/ui/OGMessageItemBody/index.tsx
+++ b/src/ui/OGMessageItemBody/index.tsx
@@ -41,7 +41,7 @@ export default function OGMessageItemBody({
   const openOGUrl = (): void => {
     let url = message?.ogMetaData?.url;
     if (url) {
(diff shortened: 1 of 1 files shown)


src/ui/OGMessageItemBody/index.tsx
```diff
@@ -37,8 +37,15 @@ export default function OGMessageItemBody({
 }: Props): ReactElement {
   const imageRef = useRef<HTMLDivElement>(null);
   const { stringSet } = useContext(LocalizationContext);
+
   const openOGUrl = (): void => {
-    if (message?.ogMetaData?.url) window.open(message?.ogMetaData?.url);
+    let url = message?.ogMetaData?.url;
+    if (url) {
+      if (!url.startsWith('http')) {
+        url = 'https://' + url;
```
liamcho, line 45: @bang9 is this proven to be safe?

src/ui/OGMessageItemBody/index.tsx
```diff
@@ -37,8 +37,15 @@ export default function OGMessageItemBody({
...
+      }
+      window.open(url);
```
liamcho, line 47: @bang9 I wanted to add try catch logic here but since this component is under `ui`, it has not senbrid provider context so we cannot import logger here. Any other suggestion you might have?

bang9: It's likely that the `window.open` won't throw an error.

src/ui/OGMessageItemBody/index.tsx
bang9: How about below?
```ts
  if (!url.startsWith('http://') && !url.startsWith('https://')) {
    url = 'https://' + url;
  }
```
I will make the same changes for react-native as well.

liamcho: Yes definitely!

liamcho: but what i meant is is adding https:// safe? is there any case that url would not work after adding https at the start?

bang9: More precisely, if site doesn't support HTTPS, it won't open.
However, for sites that have implemented at least basic handling for OG meta tags, it is expected that HTTPS would be applied lol.

bang9 (approved):
lgtm!

PR #913, opened 2024-01-11
bug-fix: Fix feedback icon button disabled

liamcho:
Fixes: [AC-992](https://sendbird.atlassian.net/browse/AC-992)

### Changelogs
- N/A

Screeshots
Before
Screenshot 2024-01-15 at 2 05 56 PM

After
Screenshot 2024-01-15 at 2 06 29 PM

[AC-992]: https://sendbird.atlassian.net/browse/AC-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Channel/components/MessageFeedbackModal/index.tsx
+++ b/src/modules/Channel/components/MessageFeedbackModal/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext, useRef, useState } from 'react';
+import React, { ReactElement, useContext, useRef } from 'react';
 import { LocalizationContext } from '../../../../lib/LocalizationContext';
 import Modal from '../../../../ui/Modal';
 import Button, { ButtonTypes } from '../../../../ui/Button';
--- a/src/modules/Channel/components/MessageFeedbackModal/index.tsx
+++ b/src/modules/Channel/components/MessageFeedbackModal/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext, useRef } from 'react';
+import React, {ReactElement, useContext, useRef, useState} from 'react';
 import { LocalizationContext } from '../../../../lib/LocalizationContext';
 import Modal from '../../../../ui/Modal';
 import Button, { ButtonTypes } from '../../../../ui/Button';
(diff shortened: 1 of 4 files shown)


src/modules/Channel/components/MessageFeedbackModal/index.tsx
```diff
@@ -30,6 +30,11 @@ export default function MessageFeedbackModal(props: MessageFeedbackModalProps):
   const { stringSet } = useContext(LocalizationContext);
   const { isMobile } = useMediaQueryContext();
 
+  const [isEdit] = useState<boolean>(
+    message?.myFeedback
+    && selectedFeedback === message.myFeedback.rating,
+  );
+
```
bang9, line 37: Is there a reason for using state?
Do we need to maintain the state as it was during the initial rendering?

liamcho: actually i wasn't sure to use state here. ill remove it since its not necessary

liamcho: done

HoonBaek (approved):
LGTM
nit: it will be good if you can attach some screen shots to show before&after

src/ui/MessageContent/index.tsx
```diff
@@ -332,6 +332,7 @@ export default function MessageContent(props: MessageContentProps): ReactElement
                     setShowFeedbackModal(true);
                   }
                 }}
+                disabled={message?.myFeedback?.rating === FeedbackRating.BAD}
```
bang9, line 335: If the rating options exceed three, it seems it might cause issues with the lower version.
What do you think about handling it this way?

`disabled={message.myFeedback && message.myFeedback.rating !== FeedbackRating.GOOD}`

liamcho: yes definitely, thanks! i applied this suggestion

bang9 (approved):
LGTM!

liamcho:
@HoonBaek 

> nit: it will be good if you can attach some screen shots to show before&after


It is done!

PR #916, opened 2024-01-11
fix: message not getting updated in real time

ishubham21:
### Issue Description
Currently, when a bot starts streaming messages, it does not reflect on the Channel UI. The reason for the same is that here we are using `message?.updatedAt` as our dependency. Now, for some reason, updatedAt is coming out to be 0, causing the dependency to have no impact. Have updated it to message to make sure it works fine. Though, the root of problem still lies in message.updatedAt coming out to be 0. 

### Replication
* Run the local storybook server
* In the Message component `src/modules/Channel/components/Message/index.tsx`, add a log to output the `message` prop. 
* Go to the integrated group channel storybook [link](https://sendbird.github.io/sendbird-uikit-react/?path=/docs/integrated-app--group-channel)
* Use the application ID and your user id to enter into the application
* Make sure one of the participants of the channel is an AI bot. 
* Give the bot some prompts
* Observe the console for message logs. You'll find that for all the message that a bot streams, the `message.updatedAt` is coming out to be 0. 

### Screenshots 
Screenshot 2024-01-11 at 7 10 08 PM

### Object

```
{
    "_iid": "**********",
    "channelType": "group",
    "messageType": "user",
    "mentionType": "users",
    "mentionedUsers": [],
    "mentionedUserIds": [],
    "metaArrays": [],
    "extendedMessage": {},
    "createdAt": 1704978949209,
    "updatedAt": 0,
    "channelUrl": "sendbird_group_channel_******f46",
    "data": "{\"metadatas\": [], \"respond_mesg_id\": 6962747609, \"not_valid\": false, \"stream\": false}",
    "customType": "",
    "mentionedMessageTemplate": "",
    "parentMessage": null,
    "silent": false,
    "isOperatorMessage": false,
    "threadInfo": null,
    "reactions": [],
    "appleCriticalAlertOptions": null,
    "scheduledInfo": null,
    "suggestedReplies": null,
    "myFeedback": null,
    "myFeedbackStatus": "NO_FEEDBACK",
    "forms": null,
    "_isContinuousMessages": false,
    "_scheduledStatus": null,
    "messageId": 6962747867,
    "parentMessageId": 0,
    "ogMetaData": null,
    "reqId": "",
    "replyToChannel": false,
    "errorCode": 0,
    "sender": {
        *****
    },
    "sendingStatus": "succeeded",
    "message": "Hello! How can I assist you today?",
    "messageParams": null,
    "translations": {},
    "translationTargetLanguages": [],
    "messageSurvivalSeconds": -1,
    "plugins": [],
    "_poll": null
}

### DIFF
--- a/src/ui/TextMessageItemBody/index.tsx
+++ b/src/ui/TextMessageItemBody/index.tsx
@@ -40,7 +40,7 @@ export default function TextMessageItemBody({
     return tokenizeMessage({
       messageText: message?.message,
     });
-  }, [message?.updatedAt]);
+  }, [message]);
   return (
     <Label
       type={LabelTypography.BODY_1}

### CODE REVIEW

File: src/ui/TextMessageItemBody/index.tsx [hunk 1]
```diff
@@ -40,7 +40,7 @@ export default function TextMessageItemBody({
     return tokenizeMessage({
       messageText: message?.message,
     });
-  }, [message?.updatedAt]);
+  }, [message]);
```
AhyoungRyu, line 43: @ishubham21 Thanks for your contribution and the detailed explanation. I see your point. 

```suggestion
  }, [message?.message]);
```
I wonder ^ this would this fix your issue too. I think this value will be updated when the streaming message is coming. 
The reason I'm asking is, putting the whole object in the dependency array is not an optimal way in terms of performance.

ishubham21: @AhyoungRyu 
Definitely this will fix the issue we are facing. This PR was just an indicate solution - the root cause lies with `updatedAt` not being set properly. 
If we proceed with this change, can we also do this for scroll events as well? I believe they too have their dependencies set as `updatedAt`

AhyoungRyu: Yeah if you're talking about [this part](https://github.com/sendbird/sendbird-uikit-react/blob/main/src/modules/Channel/components/Message/index.tsx#L152), will make the change as well.

ishubham21: @AhyoungRyu yes that’s the part. Thank you for pointing it out. Should I push the update, or sendbird team will do it?

AhyoungRyu: Oh I'll do that. Thanks for confirming it :)

AhyoungRyu: @ishubham21 Since we're not accepting the code change from outside at the moment, I made a separate PR based on yours. https://github.com/sendbird/sendbird-uikit-react/pull/928 
We're aiming to publish a new version this afternoon. We'll let you know once it's out. 

Thanks again :)

PR #923, opened 2024-01-17
fix(CLNP-1973): thread input should be disabled when it goes offline

bang9:
We don't currently support local caching for threads, so it should be disabled when it goes offline.

ticket: [CLNP-1973]

[CLNP-1973]: https://sendbird.atlassian.net/browse/CLNP-1973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HoonBaek (approved):
LGTM

PR #927, opened 2024-01-18
fix: image renderer

bang9:
- fix: fileInfoList should not be stateful to update (_This change prevents the rendering of an empty component caused by an empty array when the mfm component is initially rendered_)
- fix: scroll with scrollPubSub, scroll to bottom when change the channel
- fix: if there has been a failure in image loading when come back online, I have made modifications to retry loading the images

ticket: [CLNP-1953]

[CLNP-1953]: https://sendbird.atlassian.net/browse/CLNP-1953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Channel/context/hooks/useFileInfoListWithUploaded.ts
bang9, line 1: Since we process static data, there is no need to manage it with state.

src/modules/GroupChannel/components/MessageList/index.tsx
```diff
@@ -77,8 +78,9 @@ export const MessageList = ({
       const latestDistance = scrollDistanceFromBottomRef.current;
       const currentDistance = elem.scrollHeight - elem.scrollTop - elem.offsetHeight;
       if (latestDistance < currentDistance && (!isBottomMessageAffected || latestDistance < SCROLL_BUFFER)) {
+        const diff = currentDistance - latestDistance;
         // Move the scroll as much as the height of the message has changed
-        elem.scrollTop += currentDistance - latestDistance;
+        scrollPubSub.publish('scroll', { top: elem.scrollTop + diff, lazy: false });
```
bang9, line 83: Updates for scrolling are centrally consolidated and managed in one place.

src/modules/Channel/context/hooks/useFileInfoListWithUploaded.ts
```diff
@@ -0,0 +1,20 @@
...
+    return message.fileInfoList.map((it) => ({
+      ...it,
+      url: it.url,
+      isUploaded: true,
+    }));
+  } else {
+    return message.messageParams.fileInfoList.map((it) => ({
+      ...it,
+      url: it.fileUrl ?? (it.file instanceof Blob ? URL.createObjectURL(it.file) : undefined),
+      isUploaded: !it.file && typeof it.fileUrl === 'string' && it.fileUrl.length > 0,
```
liamcho, line 17: with this condition, there are two possible scenarios
1. file was given -> was uploaded
2. fileUrl was given -> was uploaded (not really uploaded)

if UploadedFileInfoWithUpload.isUploaded is meant for both scenarios, then no problem.

src/modules/Channel/context/hooks/useStatefulFileInfoList.ts
```diff
@@ -1,30 +0,0 @@
...
-    /**
-     * Side note: It was a bad design to not include 'isUploaded' property by SDK.
-     * Because if original object has fileUrl set and no file, then uploaded result remains
-     * the same so customer cannot know whether it has been uploaded or not.
-     */
-    (message as MultipleFilesMessage).messageParams?.fileInfoList
-      ?.map((fileInfo: UploadableFileInfo) => fileInfo.fileUrl).join(','),
-  ]);
-  return statefulFileInfoList;
-};
```
liamcho, line 30: Could you explain more what is issue with this one? 

> fix: fileInfoList should not be stateful to update (This change prevents the rendering of an empty component caused by an empty array when the mfm component is initially rendered)

src/modules/GroupChannel/context/GroupChannelProvider.tsx
```diff
@@ -291,6 +293,7 @@ const GroupChannelProvider = (props: GroupChannelContextProps) => {
         // Reset states when channel changes
         setQuoteMessage(null);
         setAnimatedMessageId(0);
+        scrollPubSub.publish('scrollToBottom', null);
```
liamcho, line 296: What if startingPoint is given? Should we not be scroll to it?

src/modules/Channel/context/hooks/useStatefulFileInfoList.ts
bang9: When processing a specific value derived from commonly existing data, there is no need to use state.

However, in exceptional cases where state needs to be used for such scenarios, you can set the initial value as follows:

```ts
useState(() => object.data.map(...))
```

But previously, the value was being populated in useEffect, causing a delay until the state value was filled.

```ts
useState([])

useEffect(() => {
  setState(object.data.map(...))
},[...])

```

src/modules/GroupChannel/context/GroupChannelProvider.tsx
bang9: This is a side effect triggered when the channelUrl is changed.

There is a separate side effect for the startingPoint below. ([link](https://github.com/sendbird/sendbird-uikit-react/blob/feat/collections/src/modules/GroupChannel/context/GroupChannelProvider.tsx#L318-L325))

Upon consideration, it seems we can execute this behavior in the clean-up of the [side effect for Scroll to bottom](https://github.com/sendbird/sendbird-uikit-react/blob/feat/collections/src/modules/GroupChannel/context/GroupChannelProvider.tsx#L298-L316).
let me change

bang9: addressed in https://github.com/sendbird/sendbird-uikit-react/pull/927/commits/48d51f9f77f74442e7bc93268989f841ec8da484

src/modules/Channel/context/hooks/useFileInfoListWithUploaded.ts
bang9: I have adopted the logic from the existing code and understood it as follows:

When a file is provided, if it is uploaded, I understand that the given file is deleted, and the fileUrl is updated.
When fileUrl is provided, since no upload is performed, We treat it as if it has been uploaded.

src/modules/Channel/context/hooks/useFileInfoListWithUploaded.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const useFileInfoListWithUploaded = (message: CoreMessageType): UploadedFileInfoWithUpload[] => {
+  if (!message || !message.isMultipleFilesMessage || !message.isMultipleFilesMessage()) return [];
+  else if (message.sendingStatus === SendingStatus.SUCCEEDED) {
+    return message.fileInfoList.map((it) => ({
+      ...it,
+      url: it.url,
+      isUploaded: true,
+    }));
+  } else {
```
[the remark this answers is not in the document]
bang9: I will add a comment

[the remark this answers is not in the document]
bang9: addressed in https://github.com/sendbird/sendbird-uikit-react/pull/927/commits/e4ad2fe5ea4e99140977466f420d2d4356d60d47

src/types.ts
```diff
@@ -74,3 +75,12 @@ export interface ClientAdminMessage extends AdminMessage, ClientMessage { }
 export interface ClientMultipleFilesMessage extends MultipleFilesMessage, ClientMessage { }
 export type EveryMessage = ClientUserMessage | ClientFileMessage | ClientMultipleFilesMessage | ClientAdminMessage;
 export type ClientSentMessages = ClientUserMessage | ClientFileMessage | ClientMultipleFilesMessage;
+
+export interface UploadedFileInfoWithUpload {
+  fileName?: string;
+  fileSize?: number;
+  mimeType?: string;
+  thumbnails?: Thumbnail[];
```
HoonBaek, line 83: How about this?

```tsx
import type { UploadedFileInfo } from '@sendbird/chat/lib/__definition';

export interface UploadedFileInfoWithUpload extends Partial<Omit<UploadedFileInfo, 'plainUrl'>> {
  isUploaded?: boolean;
}
```

src/ui/ImageRenderer/index.tsx
```diff
@@ -99,77 +87,118 @@ const ImageRenderer = ({
...
-      }
-    </div>
+      </div>
+    )
+  );
+};
+
+// Image is loaded as a background-image, but this component serves as a hidden component to receive events indicating whether the image has actually been loaded.
+const HiddenImageLoader = (props: { src: string; alt: string; onLoadStart?: () => void; onLoad?: () => void; onError?: () => void }) => {
+  const { src, alt, onLoadStart, onLoad, onError } = props;
```
HoonBaek, line 168: nit: we can use `noop` here too?

src/ui/ImageRenderer/index.tsx
```diff
@@ -99,77 +87,118 @@ const ImageRenderer = ({
...
+            setPlaceholderVisible(true);
+            setDefaultComponentVisible(false);
+          }}
+          onLoad={() => {
+            setPlaceholderVisible(false);
+            setDefaultComponentVisible(false);
+            onLoad();
+          }}
+          onError={() => {
+            setDefaultComponentVisible(true);
```
HoonBaek, line 157: nit: doesn't need to keep displaying the placehodler so...
```suggestion
            setPlaceholderVisible(true);
            setDefaultComponentVisible(true);
```

HoonBaek (approved):
LGTM

src/ui/ImageRenderer/index.tsx
```diff
@@ -99,77 +87,118 @@ const ImageRenderer = ({
...
+  const reloadCtx = useRef({
+    currSrc: src,
+    prevSrc: src,
+    loadFailure: false,
+  });
+
+  if (reloadCtx.current.currSrc !== src) {
+    reloadCtx.current.prevSrc = reloadCtx.current.currSrc;
+    reloadCtx.current.currSrc = src;
+  }
```
HoonBaek, line 179: This is a new pattern for me! 👍

src/ui/ImageRenderer/index.tsx
bang9: nice catch thank you

PR #928, opened 2024-01-19
Add message.message to message tokenize hook deps array

AhyoungRyu:
> This change is originally made by @ishubham21 at https://github.com/sendbird/sendbird-uikit-react/pull/916. Please refer to this PR for more detailed background.

`message.updateAt` won't be updated if the message is coming from a bot especially in streaming. Its value gets just `0`. I added one more field `message.message` to the dependency array so we can make sure the hook is being executed again.

### DIFF
--- a/src/modules/Thread/components/ParentMessageInfo/ParentMessageInfoItem.tsx
+++ b/src/modules/Thread/components/ParentMessageInfo/ParentMessageInfoItem.tsx
@@ -91,7 +91,7 @@ export default function ParentMessageInfoItem({
     return tokenizeMessage({
       messageText: (message as UserMessage)?.message,
     });
-  }, [message?.updatedAt]);
+  }, [message?.updatedAt, (message as UserMessage)?.message]);
 
   // Thumbnail mesage
   const [isImageRendered, setImageRendered] = useState(false);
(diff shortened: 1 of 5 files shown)


bang9 (approved):
lgtm

Do you know if this is the intended behavior on the server side or if it's a bug?

AhyoungRyu:
Yeah we've been discussed multiple times and decided not to update this `updatedAt` field. 
If the field gets updated while the message is being sent via streaming, `(edited)` will be printed too. We could use handle this side effect separately, but actually, the streaming message doesn't mean the message gets "updated". 
Not sure it sounds clear 😅

src/modules/Thread/components/ParentMessageInfo/ParentMessageInfoItem.tsx
```diff
@@ -91,7 +91,7 @@ export default function ParentMessageInfoItem({
     return tokenizeMessage({
       messageText: (message as UserMessage)?.message,
     });
-  }, [message?.updatedAt]);
+  }, [message?.updatedAt, (message as UserMessage)?.message]);
```
liamcho, line 94: is it always user message?

AhyoungRyu: Probably not. I did this for typing. See 92 line too.

liamcho: Can we please add comment so people don't get confused? same for all others below.

liamcho:
Should we add a changelog for this "bug fix"?

AhyoungRyu:
@liamcho Yes

PR #929, opened 2024-01-19
fix: handle exceptional case when channel is missing on mobile layout

bang9:
from https://github.com/sendbird/sendbird-uikit-react/pull/901#issuecomment-1899118495


AhyoungRyu (approved):
Thanks @bang9 !

PR #930, opened 2024-01-19
fix: Manage disconnected status of VoiceMessage

HoonBaek:
[CLNP-1987](https://sendbird.atlassian.net/browse/CLNP-1987)

### Issue
* Loading spinner doesn't finish if a user try to play the voice message during disconnected status

### Fix
* Reset the audio player status when failing to load the audio file with `file.fileURL`

[CLNP-1987]: https://sendbird.atlassian.net/browse/CLNP-1987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/VoicePlayer/index.tsx
+++ b/src/hooks/VoicePlayer/index.tsx
@@ -161,6 +161,13 @@ export const VoicePlayerProvider = ({
             payload: { groupKey },
           });
         };
+        audioPlayer.onerror = (error) => {
+          logger.error('VoicePlayer: Failed to load the audioFile on the audio player.', error);
+          voicePlayerDispatcher({
+            type: RESET_AUDIO_UNIT,
+            payload: { groupKey },
+          });
+        };
         audioPlayer.dataset.sbGroupId = groupKey;
         // clean up the previous audio player
         try {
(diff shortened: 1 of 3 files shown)


src/hooks/VoicePlayer/index.tsx
```diff
@@ -128,62 +129,71 @@ export const VoicePlayerProvider = ({
...
+              voicePlayerDispatcher({
+                type: ON_VOICE_PLAYER_PAUSE,
+                payload: { groupKey },
+              });
+              voicePlayerRoot.removeChild(element);
+              logger.info('VoicePlayer: Removed other player.', { element });
+            }
+          });
+        } finally {
+          audioPlayer?.play();
```
bang9, line 181: Is it okay not to set the `onerror` handler or not to use `await audio.play();`? (Handling when audio playback fails)

HoonBaek: Ah, you are right,
so the original reason is that it was failing to load the audio file, but it will be better to add error handling logic here too

Thank you for reminding me!

bang9 (approved):
LGTM!

src/hooks/VoicePlayer/index.tsx
bang9: Now it's twice as safe! LOL

PR #932, opened 2024-01-19
fix: Enable the MessageInput if cache is enabled during disconnection

HoonBaek:
[CLNP-1958](https://sendbird.atlassian.net/browse/CLNP-1958)

### Issue
* `MessageInput` and `ChannelPreview` are disabled during disconnected status even if the local caching is enabled

### Fix
* Enable the `MessageInput` and `ChannelPreview` during disconnection if the local cache is enabled

[CLNP-1958]: https://sendbird.atlassian.net/browse/CLNP-1958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/GroupChannel/context/GroupChannelProvider.tsx
+++ b/src/modules/GroupChannel/context/GroupChannelProvider.tsx
@@ -335,7 +335,7 @@ const GroupChannelProvider = (props: GroupChannelContextProps) => {
     setAnimatedMessageId(0);
     setIsScrollBottomReached(true);
 
-    if (messageDataSource.hasNext()) {
+    if (config.isOnline && messageDataSource.hasNext()) {
       await messageDataSource.resetWithStartingPoint(Number.MAX_SAFE_INTEGER);
       scrollPubSub.publish('scrollToBottom', null);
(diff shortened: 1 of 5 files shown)


bang9 (approved):
LGTM!

src/modules/GroupChannel/context/GroupChannelProvider.tsx
```diff
@@ -335,7 +335,7 @@ const GroupChannelProvider = (props: GroupChannelContextProps) => {
     setAnimatedMessageId(0);
     setIsScrollBottomReached(true);
 
-    if (messageDataSource.hasNext()) {
+    if (config.isOnline && messageDataSource.hasNext()) {
```
bang9, line 338: When the collection is initialized in offline mode, the initial value of **hasNext** is always true.
This leads to an issue of initializing the collection during the process of scrolling to the bottom after sending a message. To prevent this, I've added a check for **isOnline**

PR #937, opened 2024-01-22
bug-fix: when other user leaves, do not remove channel

aldenquimby:
- fixes a bug that may have existed for a while, or possibly have been caused by #901

### steps to reproduce:
1. User A creates a new group channel with 3+ users
2. User B leaves the channel

### expected:
- User A still sees the channel, with it's message history
- members are updated, and possibly channel avatar is updated too

### actual:
- channel immediately disappears for User A
- User A must refresh / reload the page to see the channel
- this is particularly jarring if it was your current channel and you were actively chatting

## External Contributions
- I know this PR cannot be merged directly
- I'm hoping that similar to #893 and #894 , someone on your team can pick it up


src/modules/ChannelList/dux/reducers.ts
```diff
@@ -108,28 +108,36 @@ export default function channelListReducer(
       .with({ type: channelListActions.ON_USER_LEFT }, (action) => {
         const { channel, isMe } = action.payload;
         const { allChannels, currentUserId, currentChannel, channelListQuery, disableAutoSelect } = state;
-        let nextChannels = allChannels.filter((ch) => ch.url !== channel.url);
```
aldenquimby, line 111: The diff here is worse than I'd like, and could be reduced if we want, but I found it much easier to reason about what was happening with an explicit top-level `isMe` if/else

src/modules/ChannelList/dux/reducers.ts
```diff
@@ -108,28 +108,36 @@ export default function channelListReducer(
...
+          return {
+            ...state,
+            currentChannel: nextChannel,
+            allChannels: nextChannels,
+          };
+        }
+        // If someone else left: update allChannels if needed (don't touch currentChannel)
+        else {
+          let nextChannels = allChannels.filter((ch) => ch.url !== channel.url);
+          if (!channelListQuery || filterChannelListParams(channelListQuery, channel, currentUserId)) {
```
aldenquimby, line 131: There's a change in logic here... if no channelListQuery, we should not be removing the channel (but the prior logic was incorrectly removing the channel)

liamcho:
@aldenquimby Hi thanks for providing the bug fix PR. We are onto [fixing this bug](https://github.com/sendbird/sendbird-uikit-react/pull/947) and targeted to release in the next patch version, v3.10.2.

src/modules/ChannelList/dux/reducers.ts
liamcho: Yes you are correct, thanks for spotting this out.

src/ui/ChannelAvatar/index.tsx
```diff
@@ -66,7 +66,7 @@ function ChannelAvatar({
           alt={channel?.name}
         />
       )
-  ), [channel?.members, channel?.coverUrl, theme]);
+  ), [utils.getChannelAvatarSource(channel, userId), theme]);
```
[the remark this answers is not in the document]
liamcho: Yes correct! thanks for spotting this as well! React does not recognize changes of instance properties. We will apply your sugestion as well!

liamcho:
Closing this PR because the [fix PR has been merged](https://github.com/sendbird/sendbird-uikit-react/pull/947)

aldenquimby:
Thank you @liamcho ! Appreciate it 💪

PR #938, opened 2024-01-25
bug-fix: Refactor ChannelSettings module to use channel handler for channel state update

liamcho:
Fixes: [CLNP-2099](https://sendbird.atlassian.net/browse/CLNP-2099)

### Changelogs
- Fixed a bug where muted member list is not being updated after a member had been unmuted

[CLNP-2099]: https://sendbird.atlassian.net/browse/CLNP-2099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Channel/components/MessageFeedbackModal/index.tsx
+++ b/src/modules/Channel/components/MessageFeedbackModal/index.tsx
@@ -11,11 +11,11 @@ import { FeedbackRating } from '@sendbird/chat/message';
 import { useKeyDown } from '../../../../hooks/useKeyDown/useKeyDown';
 
 export interface MessageFeedbackModalProps {
-  selectedFeedback: FeedbackRating;
+  selectedFeedback: FeedbackRating | undefined;
   message: CoreMessageType;
   onClose?: () => void;
-  onSubmit?: (comment: string) => void;
-  onUpdate?: (comment: string) => void;
+  onSubmit?: (selectedFeedback: FeedbackRating, comment: string) => void;
+  onUpdate?: (selectedFeedback: FeedbackRating, comment: string) => void;
   onRemove?: () => void;
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/ChannelSettings/components/ModerationPanel/MutedMemberList.tsx
```diff
@@ -36,16 +37,21 @@ export const MutedMemberList = (): ReactElement => {
...
-      // @ts-ignore
-      mutedMemberFilter: 'muted',
-    });
-    memberUserListQuery.next().then((members) => {
-      setMembers(members);
-      setHasNext(memberUserListQuery.hasNext);
-    });
-  }, [channel]);
+    if (channel.isSuper) {
+      const memberUserListQuery = channel?.createMemberListQuery({
```
bang9, line 41: we have `channel.createBannedUserListQuery`

bang9 (commented):
The behavior seems somewhat inconsistent.
To clearly write the code, it would be beneficial to clarify the current logic before making any modifications.

1. The current list of muted members shows only up to 10 members.
2. The list for all muted members is displayed in the MutedMemberList modal.
3. If the list of currently muted members exceeds 10 (hasNext == true), a button appears to view the MutedMemberList modal.
4. When unmuting a member, refresh the list of currently muted members:
  4-1. If unmuted from the current list.
  4-2. If unmuted from the MutedMemberList modal.
  4-3. If unmuted by another operator.
5. If a new muted member is added, refresh and display the list of muted members.

liamcho:
for 1,2, and 3, yes I will apply the limit to 10 logic. 
for 4, and 5, channel state will change which in turn triggers useEffect. This will rerender so its fine. But for super channel,this is not the case so we still need to call refreshChannel() just for super channel.

PR #942, opened 2024-01-26
fix: Apply the channelListQuery.order to the ChannelList

HoonBaek:
[CLNP-2044](https://sendbird.atlassian.net/browse/CLNP-2044)

### Fix
* Remove the action, `replace the channel to the top` in the ChannelList
* Apply the channelListQuery.order to the ChannelList

### ChangeLog
* Apply the channelListQuery.order to the ChannelList


src/modules/ChannelList/dux/reducers.ts
```diff
@@ -276,24 +276,6 @@ export default function channelListReducer(
           }),
         };
       })
-      .with({ type: channelListActions.CHANNEL_REPLACED_TO_TOP }, (action) => {
```
liamcho, line 279: why is this removed?

src/modules/ChannelList/utils.ts
```diff
@@ -302,18 +302,6 @@ export const pubSubHandler = (pubSub: SBUGlobalPubSub, channelListDispatcher: Re
     }),
   );
 
-  subscriber.set(
```
liamcho, line 305: why is this removed?

src/utils/index.ts
```diff
@@ -757,29 +757,31 @@ export const filterChannelListParams = (params: GroupChannelListQuery, channel:
...
+): Array<GroupChannel> => {
+  const compareFunc = match(order)
+    .with(GroupChannelListOrder.CHANNEL_NAME_ALPHABETICAL, () => (
+      (a: GroupChannel, b: GroupChannel) => a.name.localeCompare(b.name)
+    ))
+    .with(GroupChannelListOrder.CHRONOLOGICAL, () => (
+      (a: GroupChannel, b: GroupChannel) => b.createdAt - a.createdAt
+    ))
+    .otherwise(() => (
+      (a: GroupChannel, b: GroupChannel) => (b.lastMessage?.createdAt ?? Number.MIN_SAFE_INTEGER) - (a.lastMessage?.createdAt ?? Number.MIN_SAFE_INTEGER)
```
liamcho, line 774: What is one or both of them has no message? In that case we have to compare as CHANNEL_NAME_ALPHABETICAL

src/modules/ChannelList/dux/reducers.ts
HoonBaek: The channel item will be sorted by the `channelListQuery.order`, so we don't have to move the channel when sending a message.

same as below

src/utils/index.ts
HoonBaek: Your suggestion is reasonable, but this logic is used by server side (we can see the channel list ordered like this when fetching channel list)

HoonBaek: image
eg. The most recent channel is located last

liamcho (approved):
LG2M

PR #955, opened 2024-02-01
fix: Add missing export for the group channel modules

HoonBaek:
Add missing export for the GroupChannel and GroupChannelList modules


bang9 (approved):
lgtm

PR #963, opened 2024-02-06
fix: cannot delete property renderMessage of Object

HoonBaek:
### Fix
* Figure out error: `Cannot delete property `renderMessage` of #<Object> when using the `renderMessage` props

### Issue
* This error appears from the customer side while I try to use the props `renderMessage` of Channel
image


bang9 (commented):
```ts
const renderedMessage = useMemo(() => renderMessage?.(omitObjectProperties(props, ['renderMessage'])), [message, renderMessage]);
```

When viewed from a different perspective, are there any benefits to removing renderMessage from props? 🤔

HoonBaek:
@bang9 It looks a little bit weird that the parameter of the `renderMessage` has a `renderMessage`

bang9:
> @bang9 It looks a little bit weird that the parameter of the `renderMessage` has a `renderMessage`

Yes, so we're actually handling it with Omit in terms of types.
`type RenderMessageParamsType = Omit<MessageProps, 'renderMessage'>;`

However, in the actual code, we're not just passing `MessageProps` but passing all the values of `MessageViewProps`, so I was wondering if removing only renderMessage would make a significant difference.

bang9 (approved):

liamcho (approved):
LGTM

PR #966, opened 2024-02-08
fix: acceptableMimeTypes to support web standard format.

bang9:
- Fixed `acceptableMimeTypes` to support web standard format.


liamcho (approved):
LGTM!

PR #969, opened 2024-02-14
fix: do not disconnect and connect when the accessToken is changed

bang9:
## Description

1. When the `accessToken` is changed, `disconnect` and `connect` are called again.
2. If the customer controls the `accessToken`, the session_key refresh performed by the SDK is canceled(API request cancel) due to the `disconnect` triggered in (1).
3. Even if the `accessToken` has already expired, it is refreshed by the SDK(**SessionHandler**). Therefore, UIKit does not need to rely on this value to handle the connect logic.

ticket: [CLNP-2257]

[CLNP-2257]: https://sendbird.atlassian.net/browse/CLNP-2257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/hooks/useConnect/index.ts
+++ b/src/lib/hooks/useConnect/index.ts
@@ -46,7 +46,7 @@ export default function useConnect(triggerTypes: TriggerTypes, staticTypes: Stat
     }).catch(error => {
       logger?.error?.('SendbirdProvider | useConnect/useEffect', error);
     });
-  }, [userId, appId, accessToken]);
+  }, [userId, appId]);
   const reconnect = useCallback(async () => {
     logger?.info?.('SendbirdProvider | useConnect/reconnect/useCallback', { sdk });
 


HoonBaek (approved):
LGTM

bang9:
@HoonBaek It might also resolve [CLNP-2164], please check with the build of this branch.

HoonBaek:
@bang9 oh, wow! thank you so much!! let me check it out and resolve it

liamcho (commented):
Has this fix been tested/verified?

When SDK fails at session key refresh with below error
```
err.isSessionTokenExpiredError || err.code === SendbirdErrorCode.APPLICATION_NOT_FOUND
```
SDK awaits customer callback `onSessionTokenRequired` response to receive new auth token from customer. However, UIKit does not handle this automatically. Im not sure this is related to removing accessToekn dependency here but customer needs to handle onSessionTokenRequired outside UIKit in their own app. Am I understanding corrrecty?

liamcho:
synced offline with above question. There is no problem since UIKit opens `configureSession` such that customers can handler session events if they use accessToken.

liamcho (approved):
LGTM!

PR #970, opened 2024-02-14
fix: loading status should be handled in message list

bang9:
## Description

https://github.com/sendbird/sendbird-uikit-react/blob/v3.10.0/src/modules/Channel/components/ChannelUI/index.tsx#L59-L67
Upon reviewing the `v3.10.0` code, I noticed that the actual loading of the message list is handled in the **MessageList** component, while the loading state displayed in the `ChannelUI` relies solely on a prop called `isLoading`.

During the migration process in this area, a mistake was made. I left a note indicating the purpose of this, as I'm unsure of its intended use.

Additionally, due to the change in `isLoading` being optional, the condition for `isInvalid` became invalid in the GroupChannel module. Therefore, I added a separate fetch error to the context.

ticket: [CLNP-2259]

### DIFF
--- a/src/modules/GroupChannel/components/GroupChannelUI/GroupChannelUIView.tsx
+++ b/src/modules/GroupChannel/components/GroupChannelUI/GroupChannelUIView.tsx
@@ -37,7 +37,7 @@ export interface GroupChannelUIBasicProps {
 }
 
 export interface GroupChannelUIViewProps extends GroupChannelUIBasicProps {
-  loading: boolean;
+  isLoading?: boolean;
   isInvalid: boolean;
   channelUrl: string;
   renderChannelHeader: GroupChannelUIBasicProps['renderChannelHeader'];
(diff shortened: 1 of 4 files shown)


src/modules/GroupChannel/components/GroupChannelUI/GroupChannelUIView.tsx
```diff
@@ -62,7 +62,9 @@ export const GroupChannelUIView = (props: GroupChannelUIViewProps) => {
   const sdkError = stores?.sdkStore?.error;
   const { logger, isOnline } = config;
 
-  if (loading) {
+  // Note: This is not a loading status of the message list.
+  //  It is just for custom props from the Channel module and is not used in the GroupChannel module. (We should remove this in v4)
```
liamcho, line 66: it is included in `ChannelProps` (open public) so we should keep it as is no?

bang9: Yes, that's correct. It's being kept in ChannelUIProps

https://github.com/sendbird/sendbird-uikit-react/blob/main/src/modules/Channel/components/ChannelUI/index.tsx#L10-L12

roderickhsiao:
thanks

PR #972, opened 2024-02-14
fix: avoid numbered top-level domain

bang9:
## Description

- The regex has been modified to prevent numbers from being included in the top-level domain.
- In the Word component, the URL validation for the URL type has been removed.
- In the Word component, the use of UUID in the key has been removed.

ticket: [CLNP-2262]

[CLNP-2262]: https://sendbird.atlassian.net/browse/CLNP-2262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Message/utils/tokens/__tests__/tokenizeUtils.spec.ts
+++ b/src/modules/Message/utils/tokens/__tests__/tokenizeUtils.spec.ts
@@ -146,6 +146,26 @@ describe('identifyUrlsAndStrings', () => {
       ]);
     });
   });
+
+  it('should not parse invalid URLs', () => {
+    const invalidURLs = [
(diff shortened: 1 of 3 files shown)


HoonBaek (approved):
LGTM

AhyoungRyu (approved):

bang9:
Note: In the current change, it does not include domains consisting of English characters and numbers like Unicode. (For example, the domain `청와대.한국` is represented as `xn--vk1b187a8ue.xn--3e0b707e`)

In the future, it should be modified to detect only top-level domains composed solely of numbers.

PR #980, opened 2024-02-19
chore: Update MessageListParams

tylerhammer:
- Updated exported MessageListParams type for custom message queries


HoonBaek:
Apply in https://github.com/sendbird/sendbird-uikit-react/pull/984

PR #981, opened 2024-02-20
fix: show loading placeholder when fetching channel in ChannelSettings

bang9:
## Before
https://github.com/sendbird/sendbird-uikit-react/assets/26326015/e99d17a2-555f-4638-8d54-1182a872d4b9

## After
https://github.com/sendbird/sendbird-uikit-react/assets/26326015/5ee355b9-a0c7-497d-a857-d8c21dbaa0e6

ticket: [CLNP-2304]

[CLNP-2304]: https://sendbird.atlassian.net/browse/CLNP-2304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #983, opened 2024-02-21
refactor: update useAsyncRequest, ChannelSettingsProvider

bang9:
from https://github.com/sendbird/sendbird-uikit-react/pull/981

### DIFF
--- a/src/modules/ChannelSettings/context/ChannelSettingsProvider.tsx
+++ b/src/modules/ChannelSettings/context/ChannelSettingsProvider.tsx
@@ -48,34 +48,32 @@ interface ChannelSettingsProviderInterface {
   queries?: ChannelSettingsQueries;
   setChannelUpdateId(uniqId: string): void;
   forceUpdateUI(): void;
-  channel: GroupChannel;
+  channel: GroupChannel | null;
   loading: boolean;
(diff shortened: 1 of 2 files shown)


src/modules/ChannelSettings/context/ChannelSettingsProvider.tsx
```diff
@@ -139,7 +137,5 @@ const ChannelSettingsProvider: React.FC<ChannelSettingsContextProps> = (props: C
   );
 };
 
-export type UseChannelSettingsType = () => ChannelSettingsProviderInterface;
-const useChannelSettingsContext: UseChannelSettingsType = () => React.useContext(ChannelSettingsContext);
-
+const useChannelSettingsContext = () => React.useContext(ChannelSettingsContext);
```
HoonBaek, line 140: Doesn't need to define the types, because it sets automatically?

HoonBaek (approved):
LGTM!

src/modules/ChannelSettings/context/ChannelSettingsProvider.tsx
bang9: Yes

PR #984, opened 2024-02-21
Improve MessageListParams interface in ChannelProvider

HoonBaek:
fix from https://github.com/sendbird/sendbird-uikit-react/pull/980

thanks to @tylerhammer

### DIFF
--- a/src/modules/Channel/context/ChannelProvider.tsx
+++ b/src/modules/Channel/context/ChannelProvider.tsx
@@ -15,6 +15,7 @@ import type {
   UserMessage,
   UserMessageCreateParams,
   UserMessageUpdateParams,
+  MessageListParams as SDKMessageListParams,
 } from '@sendbird/chat/message';
 import type { EmojiContainer, SendbirdError, User } from '@sendbird/chat';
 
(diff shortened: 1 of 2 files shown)


src/modules/Channel/context/ChannelProvider.tsx
```diff
@@ -51,23 +52,13 @@ import { useHandleChannelPubsubEvents } from './hooks/useHandleChannelPubsubEven
...
-  customTypes?: Array<string>,
-  senderUserIds?: Array<string>,
-};
-
+export interface MessageListParams extends Partial<SDKMessageListParams> { // make `prevResultSize` and `nextResultSize` to optional
+  /**
+   * @deprecated
+   * It won't work even if you activate this props
+   */
+  reverse?: boolean; // TODO: Deprecate this props, because it might not work
```
bang9, line 60: todo?

bang9 (approved):

src/modules/Channel/context/ChannelProvider.tsx
HoonBaek: I thought we would delete this prop from the interface in the future

bang9: "Deprecated" indicates that the API is no longer in use and may be removed in the future.
After being deprecated, the process of removal through a major update is termed "Removed"

bang9: You have already left a **deprecated** mark~!

HoonBaek: aha;; let me remove it in the other place, thank you!

PR #992, opened 2024-02-28
feat: Add mobile layout sample to storybook

HoonBaek:


src/modules/App/stories/index.stories.js
```diff
@@ -269,6 +269,31 @@ export const Korean = () => fitPageSize(
...
+    config={{ logLevel: 'all' }}
+    replyType="THREAD"
+    isReactionEnabled
+    isMentionEnabled
+    isVoiceMessageEnabled
+    isMultipleFilesMessageEnabled
+    isMessageReceiptStatusEnabledOnChannelList
+    uikitOptions={{
+      groupChannel: {
+        // enableTypingIndicator: false,
```
AhyoungRyu, line 290: Do we want to keep this comment too?

liamcho: Let's just remove it. I left it there so we can test below option when enableTypingIndicator is set to false

liamcho (approved):
LGTM

HoonBaek:
let me close this PR, because we are preparing the new storybook of the latest version

PR #996, opened 2024-02-29
fix: Improve the emojiManager

HoonBaek:


AhyoungRyu (approved):
👍

bang9:
1. When performing such tasks, it would be good if reviewers could quantify how much improves.

2. Since the class is separated from React, it cannot detect when data loading is complete. Therefore, if you delay the loading time until the point of use by making a lazy call, there may be an issue where it is not reflected in the UI.

HoonBaek:
@AhyoungRyu 요거 닫아도 될 것 같은데 어떠신가용?

PR #1001, opened 2024-03-05
fix: add default chat view height to calculate input max height

bang9:
- When `body.offsetHeight` is 0, the height of the input is not accurately calculated.

from: https://sendbird.slack.com/archives/C0585965FFA/p1709545490650569


luke-cha (approved):
LGTM

PR #1004, opened 2024-03-08
v3.13.1

bang9:
### Fixes
- Fixed a GroupChannel scroll issue on Safari when scrollBehavior is set to smooth
- Fixed the case where the calculation of the input height in the chat window was not functioning properly


sendbird-sdk-deployment:
[Creating Ticket] 🔖 Done https://sendbird.atlassian.net/browse/SDKRLSD-1163

OnestarLee (approved):

PR #1009, opened 2024-03-13
fix: Show X button on the Modal of mobile mode

HoonBaek:
[CLNP-2528](https://sendbird.atlassian.net/browse/CLNP-2528)

### ChangeLog & Fix
* Show X button on the Modal of mobile mode

image

[CLNP-2528]: https://sendbird.atlassian.net/browse/CLNP-2528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/Modal/index.tsx
+++ b/src/ui/Modal/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, ReactNode, useContext } from 'react';
+import React, { ReactElement, ReactNode, useContext, MouseEvent } from 'react';
 import { createPortal } from 'react-dom';
 
 import './index.scss';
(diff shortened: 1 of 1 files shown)


src/ui/Modal/index.tsx
```diff
@@ -141,24 +141,20 @@ export function Modal(props: ModalProps): ReactElement {
...
+            onClick={handleClose}
+          >
+            <Icon
+              type={IconTypes.CLOSE}
+              fillColor={IconColors.DEFAULT}
+              width="24px"
+              height="24px"
+            />
+          </IconButton>
+        </div>
```
bang9, line 157: Q1. why isn't it included in header?
Q2. If there is no button available and customer were implementing custom buttons through renderHeader, what would happen?

HoonBaek: A1. It seems the ModalHeader only contains the Label(text)
A2. hm... I see,, it will be good to move the CloseButton into the ModalHeader! Then the customers' customized header won't break

Let me apply that!

HoonBaek: image
The structure has become much more stable, Thank you @bang9 !!

liamcho (approved):
LGTM

PR #1019, opened 2024-03-15
feat: Add metadataKey and metadataValues query params

aacgn:


emilbjorling:
@aacgn Hey! I'm currently fiddling around with this uikit and I've noticed that some query params aren't set properly for the GroupChannelList component as well. Maybe that should be included in this PR too?

This feels like a strange way of reporting a bug, but since the issues tab is disabled for this repo I'm doing it here 😄

chrisallo:
I close this because it's now available.

PR #1025, opened 2024-03-22
fix: Improve the ChannelSettings Modal stability

HoonBaek:
### Fix
* Improve the stability of the ChannelSettings Modals
* Fix the width size of the OGMessageItemBody component to `320px`

### DIFF
--- a/src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
+++ b/src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
@@ -29,7 +29,7 @@ export default function MembersModal({ onCancel }: Props): ReactElement {
 
   const { channel } = useChannelSettingsContext();
   const state = useSendbirdStateContext();
-  const currentUser = state?.config?.userId;
+  const currentUserId = state?.config?.userId;
   const { stringSet } = useContext(LocalizationContext);
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
```diff
@@ -67,123 +67,125 @@ export default function MembersModal({ onCancel }: Props): ReactElement {
...
-                                    };
-                                  }
-                                  return m;
-                                }));
-                                closeDropdown();
-                              }}
-                              onError={() => {
-                                // FIXME: handle error later
-                                closeDropdown();
+            members.map((member: Member) => {
```
liamcho, line 70: can you explain what has changed here?

src/ui/OGMessageItemBody/index.scss
```diff
@@ -4,9 +4,9 @@
   position: relative;
   display: inline-block;
   // TODO: Change fixing width and height lengths
-  width: 100%;
+  width: 320px;
```
liamcho, line 7: using this could result in overflow in mobile devices is this ok?

src/ui/OGMessageItemBody/index.tsx
```diff
@@ -98,6 +98,7 @@ export default function OGMessageItemBody({
           url={message?.ogMetaData?.defaultImage?.url || ''}
           alt={message?.ogMetaData?.defaultImage?.alt}
           width="100%"
+          height="280px" // FIXME: Find out better to calculate the size of OG Message
```
liamcho, line 101: using this could result in overflow in mobile devices is this ok?

src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
HoonBaek: Ah, just
```tsx
() => (<></>)
```
has changed to
```tsx
() => {
  return (<></>)
}
```
This is easier to see log, that's why

src/ui/OGMessageItemBody/index.scss
HoonBaek: We already have set the minWidth to 320px, so I think it won't be wrong for now

src/ui/OGMessageItemBody/index.tsx
HoonBaek: `height` won't cause overflow, actually this reduced the height on the mobile

src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
```diff
@@ -29,7 +29,7 @@ export default function MembersModal({ onCancel }: Props): ReactElement {
 
   const { channel } = useChannelSettingsContext();
   const state = useSendbirdStateContext();
-  const currentUser = state?.config?.userId;
+  const currentUserId = state?.config?.userId;
```
bang9, line 32: 👍따봉

liamcho (approved):
LGTM

PR #1026, opened 2024-03-22
fix: Replace the root components order

HoonBaek:
### Fix
* Replace the root components order because dropdown menu is hided by modals

before
image

after
image


liamcho (approved):
LGTM

PR #1032, opened 2024-03-28
feat: Remove carousel component and use message template instead + some spec changes

liamcho:
Fixes: [AC-1230](https://sendbird.atlassian.net/browse/AC-1230)

- 카루셀을 UIKit 에서 빼고 Core 쪽으로 옮긴다.
- 카루셀은 View style 무시, container type 무시 (미래에 요청 있을 때 오픈)
- container type full 은 뺀다 => default, wide 만 지원한다 => 지원 없는 container type 을 받으면 default 로 그려준다.
- template version validation 을 추가한다.

### Changelogs
- Removed `Carousel` ui component as message template returns carousel component instead
- Container type 'full' is no longer supported due to inconsistent default style of parent template components with other types
- Added template version validation. If template's version is unsupported, fallback message will be rendered instead.

[AC-1230]: https://sendbird.atlassian.net/browse/AC-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ui/TemplateMessageItemBody/index.tsx
```diff
@@ -192,10 +190,14 @@ export function TemplateMessageItemBody({
             logger.error('TemplateMessageItemBody | no reservation key found in view_variables: ', reservationKey, templateData.view_variables);
             throw new Error();
           }
-          result.filledMessageTemplateItemsList = getFilledMessageTemplateItemsForCarouselTemplate(
-            simpleTemplateDataList,
-          );
-          result.carouselItem = carouselItem;
+          result.templateVersion = cachedTemplate.version;
+          result.filledMessageTemplateItemsList = [{
+            type: carouselItem.type as any,
```
AhyoungRyu, line 195: Do we have reason for converting its type to any here? I think it should be fine without this.

AhyoungRyu (approved):
LGTM. Please check the comment I left before you merge :)

src/ui/TemplateMessageItemBody/index.tsx
liamcho: No input is string but the type is a enum defined in the core package. The type is not exported so i could not import. I had to cast to any.

PR #1047, opened 2024-04-09
fix: Support optional customize functions returning null case

liamcho:
Fixes: [CLNP-2851](https://sendbird.atlassian.net/browse/CLNP-2851)

### Changelogs
- Fixed a bug where injecting an optional property with null value not rendering the expected default component

[CLNP-2851]: https://sendbird.atlassian.net/browse/CLNP-2851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/ChannelSettings/components/LeaveChannel/index.tsx
+++ b/src/modules/ChannelSettings/components/LeaveChannel/index.tsx
@@ -5,7 +5,7 @@ import type { GroupChannel } from '@sendbird/chat/groupChannel';
 
 import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
 import { useChannelSettingsContext } from '../../context/ChannelSettingsProvider';
-import { noop } from '../../../../utils/utils';
+import { deleteNullish, noop } from '../../../../utils/utils';
 
 import Modal from '../../../../ui/Modal';
 import { useLocalization } from '../../../../lib/LocalizationContext';
@@ -25,7 +25,7 @@ const LeaveChannel: React.FC<LeaveChannelProps> = (props: LeaveChannelProps) => 
   const {
     onSubmit = noop,
     onCancel = noop,
-  } = props;
+  } = deleteNullish(props);
 
   const { channel, onLeaveChannel } = useChannelSettingsContext();
   const { stringSet } = useLocalization();
--- a/src/modules/ChannelSettings/components/LeaveChannel/index.tsx
+++ b/src/modules/ChannelSettings/components/LeaveChannel/index.tsx
@@ -5,7 +5,7 @@ import type { GroupChannel } from '@sendbird/chat/groupChannel';
 
 import useSendbirdStateContext from '../../../../hooks/useSendbirdStateContext';
 import { useChannelSettingsContext } from '../../context/ChannelSettingsProvider';
-import { deleteNullish, noop } from '../../../../utils/utils';
+import { noop } from '../../../../utils/utils';
 
 import Modal from '../../../../ui/Modal';
 import { useLocalization } from '../../../../lib/LocalizationContext';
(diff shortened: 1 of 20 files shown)


bang9:
TODO:

- [ ] src/modules/ChannelSettings/components/ChannelSettingsUI/index.tsx
- [ ] src/modules/Channel/components/ChannelUI/index.tsx 
- [ ] src/modules/Channel/components/MessageList/index.tsx
- [ ] src/modules/GroupChannel/components/Message/MessageView.tsx
- [ ] src/modules/GroupChannel/components/MessageList/index.tsx
- [ ] src/ui/MessageContent/index.tsx
- [ ] src/ui/MessageItemMenu/index.tsx

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/FeedbackIconButton/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/IconButton/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/ImageRenderer/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/MessageInput/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/Modal/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/ReactionBadge/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/ui/Toggle/ToggleUI.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/modules/ChannelSettings/components/LeaveChannel/index.tsx
bang9, line 1: It doesn't seem to be the target that needs modification

src/modules/GroupChannel/components/Message/MessageView.tsx
```diff
@@ -146,7 +147,7 @@ const MessageView = (props: MessageViewProps) => {
     animatedMessageId,
     onMessageAnimated,
     usedInLegacy = true,
-  } = props;
+  } = deleteNullish(props);
```
bang9, line 150: @liamcho @HoonBaek @AhyoungRyu

When `renderXXX` props are mixed with regular props like this, would it be better to separate them?
Have concerned about potential side effects in the logic when customers pass `null` and it changes to `undefined`. (For example, when business logic checks conditions using `=== null`.)

```tsx
// AS-IS
const {
  propA,
  propB,
  renderPropC = () => <DefaultComponent />
} = deleteNullish(props);

// TO-BE
const { propA, propB } = props;
const { renderPropC = () => <DefaultComponent /> } = deleteNullish(props);
```

liamcho: Its the same no? propA, propB are references so they could be removed after deleteNullish(props)

liamcho (commented):
@bang9 can you commit directly and merge while im away?

src/modules/GroupChannel/components/Message/MessageView.tsx
bang9: If we are distinguishing between undefined and null for propA and propB in our business logic, the outcome will differ. (The customer passed null, but it is handled as undefined)

HoonBaek (approved):
LGTM

Is it intentional that `deleteNullish` is separated in some parts and not in others?

bang9:
> LGTM
> 
> Is it intentional that `deleteNullish` is separated in some parts and not in others?

Yes, you can read more -> https://github.com/sendbird/sendbird-uikit-react/pull/1047#discussion_r1558840216

PR #1055, opened 2024-04-17
fix: connect the app props and channenl list props

HoonBaek:
### Fix & ChangeLog
* connect the app props to the group channel list
  * App: `isTypingIndicatorEnabledOnChannelList` and `isMessageReceiptStatusEnabledOnChannelList`

### DIFF
--- a/src/modules/App/DesktopLayout.tsx
+++ b/src/modules/App/DesktopLayout.tsx
@@ -12,6 +12,7 @@ import ChannelSettings from '../ChannelSettings';
 import MessageSearchPannel from '../MessageSearch';
 import Thread from '../Thread';
 import { SendableMessageType } from '../../utils';
+import useSendbirdStateContext from '../../hooks/useSendbirdStateContext';
 
 export const DesktopLayout: React.FC<DesktopLayoutProps> = (props: DesktopLayoutProps) => {
   const {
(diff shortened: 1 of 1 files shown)


src/modules/App/DesktopLayout.tsx
```diff
@@ -70,6 +76,8 @@ export const DesktopLayout: React.FC<DesktopLayoutProps> = (props: DesktopLayout
     selectedChannelUrl: currentChannel?.url,
     onChannelCreated: updateFocusedChannel,
     onUserProfileUpdated: onProfileEditSuccess,
+    isTypingIndicatorEnabled: isTypingIndicatorEnabledOnChannelList,
+    isMessageReceiptStatusEnabled: isMessageReceiptStatusEnabledOnChannelList,
```
bang9, line 80: Maybe.. the global-level configs shouldn't be injected into the module-level settings

HoonBaek: This is a DesktopLayout component, and it should accept the global level options, if not it won't work as our expect

So, the current issue is that the two options `isTypingIndicatorEnabledOnChannelList` and `isMessageReceiptStatusEnabledOnChannelList` are not making any impact.

bang9: It sounds like the module isn't adhering to those global settings. Could you please confirm that part?

PR #1065, opened 2024-04-18
release: v3.14.2

HoonBaek:
### Fixes
* Fixed a bug where right padding is added to messages sent by me in mobile devices
* Removed image section in the OGMessageItemBody if there is no og image
* Fixed that safely opens URL to prevent XSS
* Fixed that copying URI-list issue in the iOS device/Safari
* Fixed that channel badge count is not updated on iOS Webview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,6 @@
 
 ## [v3.14.2] (Apr 18, 2024)
 
-### Features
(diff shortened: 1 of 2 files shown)


sendbird-sdk-deployment:
[Creating Ticket] 🔖 Done https://sendbird.atlassian.net/browse/SDKRLSD-1238

PR #1080, opened 2024-04-29
feat: docs

bang9:
- Added typedoc and `build:docs` script

ticket: [CLNP-2426]

[CLNP-2426]: https://sendbird.atlassian.net/browse/CLNP-2426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


typedoc.json
HoonBaek, line 1: If we don't add the new file to the `typedoc.json` file when creating a new file, does that mean the corresponding type doc won't be generated?

In that case, adding instructions in files like README to include the new file in the `typedoc.json` file when creating it would be helpful.

HoonBaek (approved):

typedoc.json
bang9: All modules exported from `rollup.module-exports.mjs` are automatically updated to the typedoc.json.
No additional processing is required.


https://github.com/sendbird/sendbird-uikit-react/pull/1080/files#diff-2eae17131daff87ab901db02dbfd2a57881eac8c3fdcd8ac82dcb71bcc0bac75R3

PR #1085, opened 2024-05-02
fix: Move import statements to the top of extracted index.css file.

liamcho:
Fixes: [CLNP-3084](https://sendbird.atlassian.net/browse/CLNP-3084)

### Changelogs
- Fixed a bug where import statements are not located at the top of the extracted index.css file

[CLNP-3084]: https://sendbird.atlassian.net/browse/CLNP-3084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -58,7 +58,7 @@ export default {
           resolvecss({ code: result.css.toString() });
         }),
       plugins: [autoprefixer],
-      sourceMap: true,
+      sourceMap: false,
       extract: "dist/index.css",
       extensions: [".sass", ".scss", ".css"],
     }),
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -14,6 +14,7 @@ import ts2 from "rollup-plugin-typescript2"
 // config from package.json
 import pkg from "./package.json" assert {type: "json"};
 import inputs from "./rollup.module-exports.mjs";
+import {readFileSync, writeFileSync} from 'fs';
 
 const APP_VERSION_STRING = "__react_dev_mode__";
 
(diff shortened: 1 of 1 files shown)


rollup.config.mjs
```diff
@@ -14,6 +14,7 @@ import ts2 from "rollup-plugin-typescript2"
 // config from package.json
 import pkg from "./package.json" assert {type: "json"};
 import inputs from "./rollup.module-exports.mjs";
+import {readFileSync, writeFileSync} from 'fs';
```
bang9, line 17:
```suggestion
import { readFileSync, writeFileSync } from 'fs';
```

rollup.config.mjs
```diff
@@ -61,6 +62,41 @@ export default {
       extract: "dist/index.css",
       extensions: [".sass", ".scss", ".css"],
     }),
+    {
```
bang9, line 65: When you compile Sass, it generates a CSS map (sourcemap) file that helps to track the positions of CSS and original Sass files. However, if the order of text in the CSS file changes, the sourcemap may no longer be valid.

Regarding whether there will be any impact on this issue (whether it's safe to remove it), personally, I think setting sourcemap: false shouldn't cause significant problems.


```js
postcss({
  // ...
  sourceMap: false,
})
```

bang9: Hmm.. let's remove it

bang9:
could you update the title ? `bug-fix` -> `fix`
reference: https://www.conventionalcommits.org/en/v1.0.0/

rollup.config.mjs
liamcho: interesting. Ill look into this maybe next week due to other priorities

liamcho: @bang9 I asked chatGPT and below is answer

> If you add a custom plugin like 'move-imports-to-top', it won't automatically handle sourcemaps for you. Sourcemaps are typically handled by the plugins provided by Rollup or other build tools.

It also suggested me to handle mapping manually using
`import { SourceMapConsumer, SourceMapGenerator } from 'source-map';`

Do you think we should try this? or just set sourceMap: false?

bang9: `sourceMap: false` let's go~

bang9 (approved):

PR #1088, opened 2024-05-04
v3.14.5

liamcho:
### Fixes
* Fixed a bug where channel scroll to bottom is not called internally when last message is updated with suggested replies


sendbird-sdk-deployment:
[Creating Ticket] 🔖 Done https://sendbird.atlassian.net/browse/SDKRLSD-1272

bang9 (approved):

PR #1094, opened 2024-05-17
refactor: stable scroll behavior

bang9:
## Changes
- Improvement of scroll position flickering when load previous messages
- Attempt to load based on screen size threshold

## test link
before: https://sendbird-uikit-react.netlify.app/
after: https://deploy-preview-1094--sendbird-uikit-react.netlify.app/

ticket: [CLNP-3273]

[CLNP-3273]: https://sendbird.atlassian.net/browse/CLNP-3273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/GroupChannel/context/GroupChannelProvider.tsx
```diff
@@ -204,60 +200,6 @@ export const GroupChannelProvider = (props: GroupChannelProviderProps) => {
...
-
-      preventDuplicateRequest.release();
-
-      if (currentChannel && !messageDataSource.hasNext()) {
-        messageDataSource.resetNewMessages();
-        if (!disableMarkAsRead) markAsReadScheduler.push(currentChannel);
-      }
-    },
-  });
-
```
bang9, line 260: Moved to `<InfiniteList />`

src/modules/GroupChannel/context/hooks/useMessageListScroll.tsx
```diff
@@ -102,27 +92,27 @@ export function useMessageListScroll(behavior: 'smooth' | 'auto') {
...
-    onInBetween({ distanceFromBottom }) {
-      setIsScrollBottomReached(false);
-      scrollDistanceFromBottomRef.current = distanceFromBottom;
-    },
-    onReachedBottom({ distanceFromBottom }) {
-      setIsScrollBottomReached(true);
-      scrollDistanceFromBottomRef.current = distanceFromBottom;
-    },
-  });
-
```
bang9, line 120: Moved to `<InfiniteList />`

src/modules/GroupChannel/context/hooks/useMessageListScroll.tsx
```diff
@@ -25,41 +24,32 @@ export type ScrollTopicUnion =
...
-export function useMessageListScroll(behavior: 'smooth' | 'auto') {
+export function useMessageListScroll(behavior: 'smooth' | 'auto', deps: DependencyList = []) {
   const scrollRef = useRef<HTMLDivElement>(null);
+  const scrollPositionRef = useRef(0);
   const scrollDistanceFromBottomRef = useRef(0);
 
-  const [scrollPubSub] = useState(() => pubSubFactory<ScrollTopics, ScrollTopicUnion>());
-  const [isScrollBottomReached, setIsScrollBottomReached] = useState(false);
+  const [scrollPubSub] = useState(() => pubSubFactory<ScrollTopics, ScrollTopicUnion>({ publishSynchronous: true }));
```
AhyoungRyu, line 32: Any reason we use useState here instead of useRef? It seems it doesn't have a set function.

src/modules/GroupChannel/context/GroupChannelProvider.tsx
```diff
@@ -152,7 +150,7 @@ export const GroupChannelProvider = (props: GroupChannelProviderProps) => {
   const [fetchChannelError, setFetchChannelError] = useState<SendbirdError | null>(null);
 
   // Ref
-  const { scrollRef, scrollPubSub, scrollDistanceFromBottomRef, isScrollBottomReached, setIsScrollBottomReached } = useMessageListScroll(scrollBehavior);
+  const { scrollRef, scrollPubSub, scrollDistanceFromBottomRef, isScrollBottomReached, setIsScrollBottomReached, scrollPositionRef } = useMessageListScroll(scrollBehavior, [currentChannel]);
```
AhyoungRyu, line 153:
```suggestion
  const { scrollRef, scrollPubSub, scrollDistanceFromBottomRef, isScrollBottomReached, setIsScrollBottomReached, scrollPositionRef } = useMessageListScroll(scrollBehavior, [currentChannel?.url);
```
I think having a url(simple string) would be enough for this instead of putting an huge object.

src/lib/pubSub/index.ts
```diff
@@ -42,7 +46,11 @@ const pubSubFactory = <
     publish: (topic, info) => {
       if (topics[topic]) {
         topics[topic].forEach((fn) => {
-          setTimeout(() => fn(info !== undefined ? info : {}), 0);
+          if (opts?.publishSynchronous) {
```
HoonBaek, line 49: What's the purpose of synchronous?

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { forwardRef, UIEventHandler, useLayoutEffect, useRef } from 'react';
+import { BaseMessage } from '@sendbird/chat/message';
```
HoonBaek, line 2: nit: ```import type { BaseMessage } from '@sendbird/chat/message';```

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
```diff
@@ -0,0 +1,104 @@
...
+      direction.current = 'bottom';
+      await onLoadNext();
+      isFetching.current = false;
+    } else {
+      direction.current = undefined;
+    }
+  };
+
+  return (
+    <div className="sendbird-conversation__scroll-container">
```
HoonBaek, line 84: isn't it better to name it `sendbird-inifinite-list` ? or adding it, if you considered the className legacy

HoonBaek (approved):
LGTM

src/lib/pubSub/index.ts
bang9: When publishing events, you can choose between synchronous and asynchronous execution.
Here are the differences illustrated with examples:

## Synchronous Execution:

```js
fns.forEach(fn => fn());
```
- Blocking: Each function runs to completion before the next one starts, potentially causing delays if any function takes time.
- Predictable Order: Functions execute in the exact order they are called.

## Asynchronous Execution:
```js
fns.forEach(fn => setTimeout(() => fn(), 0));
```
- Non-blocking: Functions are scheduled to run after the current execution context, keeping the main thread free.
- Less Predictable Order: Functions might not execute in the strict order they are scheduled due to other asynchronous tasks.

src/modules/GroupChannel/context/GroupChannelProvider.tsx
bang9: -> https://github.com/sendbird/sendbird-uikit-react/pull/1094/commits/0d9127c226fa19e1cc6b513c2183335cca38537f

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
bang9: As an abstracted component for logic only, not UI, I named it somewhat arbitrarily since all the names like MessageList were already taken.

Currently, I don't know how it will be improved in the future, so I don't want to create backward compatibility issues.

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
bang9: -> https://github.com/sendbird/sendbird-uikit-react/commit/0d9127c226fa19e1cc6b513c2183335cca38537f

AhyoungRyu (approved):

PR #1095, opened 2024-05-18
fix: mention

bang9:
- Fix the issue where the mention feature does not work properly in the input component
- Fix the issue where unnecessary spaces are added between mention texts when editing an already mentioned message


chrisallo (approved):
lgtm!

PR #1105, opened 2024-05-23
fix: scroll to the bottom of the list when mounted before painting

bang9:
- scroll to bottom is not working properly in the widget after applying the list refactoring

https://github.com/sendbird/sendbird-uikit-react/assets/26326015/bb5c3d5e-923e-4ec2-864a-62905d8a2558


HoonBaek (approved):
LGTM

PR #1106, opened 2024-05-24
chore: Add missing margin to suggested replies

liamcho:
ref:
https://sendbird.slack.com/archives/C0585965FFA/p1716533846284719?thread_ts=1716532857.466669&cid=C0585965FFA


src/modules/GroupChannel/components/SuggestedReplies/index.tsx
```diff
@@ -1,15 +1,36 @@
...
+export const ReplyItem = ({
+  value,
+  index,
+  onClickReply,
+}: ReplyItemProps) => {
+  return (
+    <div
+      className="sendbird-suggested-replies__option"
+      id={value}
+      key={index + value}
```
AhyoungRyu, line 26:
```suggestion
  onClickReply,
}: ReplyItemProps) => {
  return (
    <div
      className="sendbird-suggested-replies__option"
      id={value}
```
We don't need the key here since you pass it inside of the children below :)

bang9:
please resolve build error

src/modules/GroupChannel/components/SuggestedReplies/index.tsx
bang9: `index` can be removed from the props as well

PR #1107, opened 2024-05-24
fix: retry connect when failed with token expired error

bang9:
## Changes
- When attempting to `connect` with an expired or invalid session token, an error will be returned.
In such cases, the error message is checked, and if a `SessionHandler` is available, it will extract a valid session token and retry the `connect`.
- The `eventHandlers.connection.onFailed(SendbirdError)` will not be called when userId and appId are missing.
This should not be triggered as it is not related to connection errors. (In the first place, this logic should not be reached without an app ID and user ID.)
- Added `eventHandlers.connection.onConnected(User)`

## Recording
https://github.com/sendbird/sendbird-uikit-react/assets/26326015/ab529a97-7fe8-4e47-aa2c-9047886f22c8

- ticket: [AC-2476]

[AC-2476]: https://sendbird.atlassian.net/browse/AC-2476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/hooks/useConnect/__test__/setupConnection.spec.ts
+++ b/src/lib/hooks/useConnect/__test__/setupConnection.spec.ts
@@ -2,23 +2,19 @@
 /* eslint-disable global-require */
 import { SDK_ACTIONS } from '../../../dux/sdk/actionTypes';
 import { USER_ACTIONS } from '../../../dux/user/actionTypes';
-import { getMissingParamError, setUpConnection, setUpParams } from '../setupConnection';
+import { getMissingParamError, setUpConnection, initSDK, getConnectSbError } from '../setupConnection';
 import { SetupConnectionTypes } from '../types';
 import { generateSetUpConnectionParams, mockSdk, mockUser, mockUser2 } from './data.mocks';
-
 import { SendbirdError } from '@sendbird/chat';
 
 // todo: combine after mock-sdk is implemented
-jest.mock('@sendbird/chat', () => ({
-  init: jest.fn().mockImplementation(() => mockSdk),
-  SendbirdError: jest.fn(),
-}));
-jest.mock('@sendbird/chat/openChannel', () => ({
-  OpenChannelModule: jest.fn(),
-}));
-jest.mock('@sendbird/chat/groupChannel', () => ({
-  GroupChannelModule: jest.fn(),
-}));
+jest.mock('@sendbird/chat', () => {
+  const originalModule = jest.requireActual('@sendbird/chat');
+  return {
+    init: jest.fn(() => mockSdk),
+    ...originalModule,
+  };
+});
 
 describe('useConnect/setupConnection', () => {
   it('should call SDK_ERROR when there is no appId', async () => {
@@ -26,8 +22,7 @@ describe('useConnect/setupConnection', () => {
     const params = { ...setUpConnectionProps, appId: undefined };
     const errorMessage = getMissingParamError({ userId: params.userId, appId: params.appId });
 
-    await expect(setUpConnection((params as unknown as SetupConnectionTypes)))
-      .rejects.toMatch(errorMessage);
+    await expect(setUpConnection(params as unknown as SetupConnectionTypes)).rejects.toMatch(errorMessage);
 
     expect(mockSdk.connect).not.toBeCalledWith({ type: SDK_ACTIONS.SET_SDK_LOADING, payload: true });
     expect(setUpConnectionProps.sdkDispatcher).toBeCalledWith({ type: SDK_ACTIONS.SDK_ERROR });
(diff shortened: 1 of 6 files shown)


src/lib/hooks/useConnect/setupConnection.ts
bang9, line 1: I would recommend check out instead of diff.

AhyoungRyu (approved):
- It'b be better we can separate onConnected & onConnectFailed from setUpConnection as well since it's still huge. 
- It seems `Steps` comment needs to be updated as well 

Quick two comments from my side and not they're not blockers :)

bang9:
> * It'b be better we can separate onConnected & onConnectFailed from setUpConnection as well since it's still huge.
> * It seems `Steps` comment needs to be updated as well
> 
> Quick two comments from my side and not they're not blockers :)

@AhyoungRyu @liamcho @HoonBaek
I’m considering removing the `Steps` comments. What do you think?
I don’t think the logic is difficult enough to require an explanation.

AhyoungRyu:
@bang9 Yeah I think it's okay to remove.

src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -82,126 +66,190 @@ export async function setUpConnection({
...
-        .catch((err) => connectCbError(err));
     } else {
       const errorMessage = getMissingParamError({ userId, appId });
+      const error = new Error(errorMessage);
+      logger.error?.(error.message);
       sdkDispatcher({ type: SDK_ERROR });
-      onConnectionFailed?.({ message: errorMessage } as SendbirdError);
-      logger?.error?.(errorMessage);
-      // exit promise with error
+
```
liamcho, line 170: minor: We could move this to the top for readability
```
      if (!userId || !appId) {
        const errorMessage = getMissingParamError({ userId, appId });
        const error = new Error(errorMessage);
        logger.error?.(error.message);
        sdkDispatcher({ type: SDK_ERROR });

        reject(errorMessage);
      }
```

src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -82,126 +66,190 @@ export async function setUpConnection({
...
+        logger.error?.(errorMessage, { e, appId, userId });
         userDispatcher({ type: RESET_USER });
-
+        sdkDispatcher({ type: RESET_SDK });
         sdkDispatcher({ type: SDK_ERROR });
-        onConnectionFailed?.(e);
-        // exit promise with error
+
         reject(errorMessage);
+        eventHandlers?.connection?.onFailed?.(e);
```
liamcho, line 163: shouldn't reject be called at the last? or it does not matter?

src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -82,126 +66,190 @@ export async function setUpConnection({
...
-              nickname,
-              profileUrl,
-            });
-            userDispatcher({ type: UPDATE_USER_INFO, payload: namedUser });
-          }).finally(() => {
-            resolve();
-          });
-        } else {
-          resolve();
+        try {
```
liamcho, line 126: TODO: better to take wrap this logic in the future

src/lib/hooks/useConnect/__test__/setupConnection.spec.ts
```diff
@@ -2,32 +2,27 @@
...
-  SendbirdError: jest.fn(),
-}));
-jest.mock('@sendbird/chat/openChannel', () => ({
-  OpenChannelModule: jest.fn(),
-}));
-jest.mock('@sendbird/chat/groupChannel', () => ({
-  GroupChannelModule: jest.fn(),
-}));
+jest.mock('@sendbird/chat', () => {
+  const originalModule = jest.requireActual('@sendbird/chat');
```
liamcho, line 12: why did we have to change to requireActual?

src/lib/hooks/useConnect/setupConnection.ts
```diff
@@ -82,126 +66,190 @@ export async function setUpConnection({
...
+          // NOTE: The part that connects via the SDK must be callable directly by the customer.
+          //  we should refactor this in next major version.
+          if (shouldRetryWithValidSessionToken(error) && sessionHandler) {
+            try {
+              const sessionToken = await new Promise(sessionHandler.onSessionTokenRequired);
+              if (sessionToken) {
+                logger.info?.(
+                  `SendbirdProvider | useConnect/setupConnection/connect retry connect with valid session token: ${sessionToken.slice(0, 10) + '...'}`,
+                );
+                const user = await sdk.connect(userId, sessionToken);
```
liamcho, line 92: Nice! so this logic was missing before

liamcho (approved):
LGTM!

src/lib/hooks/useConnect/__test__/setupConnection.spec.ts
bang9: If the rest of the module is not re-added, only the mocked method will be returned. (If there is code that uses the rest of the module, an undefined error will occur.)

bang9: If such an error occurs, the test code needs to be modified.
Previously, the actual code included logic to avoid this error(`SendbirdProduct?.UIKIT_CHAT ?? 'uikit-chat'`)
which made the code somewhat complex.

src/lib/hooks/useConnect/setupConnection.ts
bang9: It doesn’t really matter

PR #1108, opened 2024-05-25
fix: add workaround to reset IME in mobile webkit

bang9:
## Description
contentEditable does not work as expected in mobile WebKit (Safari).
Events and properties related to composing, necessary for combining characters like Hangul, do not seem to be passed to IME properly.

When resetting the input with innerHTML, the previous context should be cleared.
However, if the last text of the previous input was a composable string, it appears in the next input.

1. If the last string is not a composable string, it is not affected. This is why preventing the `e.preventDefault()` call when sending a message with Enter prevents the bug, as the Enter (newline) string is the last text in the IME context and is not affected by this bug.
2. The bug can be avoided by resetting the IME context(blur or moving caret) of the last string. This is similar to how moving the cursor while typing Hangul breaks the composition.

## Changes
In this PR, an attempt is made to reset by moving keyboard focus to another input, and it operates as follows:

1. A ghost input, which is not visible in the UI but selectable via focus, is created.
2. After sending the text and resetting with innerHTML, focus immediately shifts to the ghost input:
   - To prevent the iPhone's virtual keyboard from closing:
   - Changing keyboard focus causes the loss of context for the composing IME string.
   - Using `requestAnimationFrame` to move focus back to the original input:
     - To prevent the change of focus from being treated as input → ghost input → input too quickly, without allowing a minimum interval.
     - Using `setTimeout` or similar causes delays due to waiting for the event loop, resulting in the phenomenon of the virtual keyboard going up and down. Therefore, `requestAnimationFrame`, associated with the browser's cycle of drawing new frames, is used.
3. Focus is then returned to the original input.

## Restrictions
The above approach only applies to input from the `Virtual Keyboard` (like an iPhone keyboard).
The issue still persists when typing using a `Keyboard` in the mobile device.

ticket: [CLNP-3470]

[CLNP-3470]: https://sendbird.atlassian.net/browse/CLNP-3470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ui/MessageInput/index.tsx
+++ b/src/ui/MessageInput/index.tsx
@@ -133,6 +133,7 @@ const MessageInput = React.forwardRef<HTMLInputElement, MessageInputProps>((prop
   } = props;
 
   const internalRef = (externalRef && 'current' in externalRef) ? externalRef : null;
+  const ghostInputRef = useRef<HTMLInputElement>(null);
 
   const textFieldId = messageFieldId || TEXT_FIELD_ID;
   const { stringSet } = useLocalization();
@@ -394,8 +395,19 @@ const MessageInput = React.forwardRef<HTMLInputElement, MessageInputProps>((prop
       const params = { message: messageText, mentionTemplate };
       onSendMessage(params);
       resetInput(internalRef);
-      // important: keeps the keyboard open -> must add test on refactor
-      textField.focus();
+
+      /**
+       * Note: contentEditable does not work as expected in mobile WebKit (Safari).
+       * @see https://github.com/sendbird/sendbird-uikit-react/pull/1108
+       */
+      if (isMobileIOS(navigator.userAgent)) {
+        if (ghostInputRef.current) ghostInputRef.current.focus();
+        requestAnimationFrame(() => textField.focus());
+      } else {
+        // important: keeps the keyboard open -> must add test on refactor
+        textField.focus();
+      }
+
       setIsInput(false);
       setHeight();
     }
@@ -426,6 +438,14 @@ const MessageInput = React.forwardRef<HTMLInputElement, MessageInputProps>((prop
       disabled && 'sendbird-message-input-form__disabled',
     )}>
       <div className={classnames('sendbird-message-input', disabled && 'sendbird-message-input__disabled')} data-testid="sendbird-message-input">
+        {isMobileIOS(navigator.userAgent) && (
+          <input
+            id={'ghost-input-reset-ime-cjk'}
+            ref={ghostInputRef}
+            style={{ opacity: 0, padding: 0, margin: 0, height: 0, border: 'none', position: 'absolute', top: -9999 }}
+            defaultValue={'_'}
+          />
+        )}
         <div
           id={`${textFieldId}${isEdit ? message?.messageId : ''}`}
           className={`sendbird-message-input--textarea ${textFieldId}`}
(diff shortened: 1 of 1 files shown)


bang9:
After exploring the root cause further, we found that when a composable string is typed, the following events should be triggered:
[oncompositionstart](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event) / [oncompositionupdate](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionupdate_event) / [oncompositionend](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event).

If a non-composable string, such as Enter, is typed, oncompositionend is triggered. Even if an action like setting innerHTML to initialize occurs, oncompositionend is still called. After that, as expected, text composition should no longer occur, and the next input, even if it is a combinable character, should be processed as individual inputs.

However, it seems that mobile Safari does not handle this correctly. To verify this, tests were conducted, but while typing Korean in mobile Safari, these events could not be captured.

bang9:
https://bugs.webkit.org/show_bug.cgi?id=274702
I have submitted the issue.

src/ui/MessageInput/index.tsx
```diff
@@ -426,6 +438,14 @@ const MessageInput = React.forwardRef<HTMLInputElement, MessageInputProps>((prop
       disabled && 'sendbird-message-input-form__disabled',
     )}>
       <div className={classnames('sendbird-message-input', disabled && 'sendbird-message-input__disabled')} data-testid="sendbird-message-input">
+        {isMobileIOS(navigator.userAgent) && (
+          <input
+            id={'ghost-input-reset-ime-cjk'}
```
AhyoungRyu, line 443: Oh cjk means; Chinese Japanese Korean 😂

AhyoungRyu (approved):

src/ui/MessageInput/index.tsx
```diff
@@ -394,8 +395,19 @@ const MessageInput = React.forwardRef<HTMLInputElement, MessageInputProps>((prop
...
-      // important: keeps the keyboard open -> must add test on refactor
-      textField.focus();
+
+      /**
+       * Note: contentEditable does not work as expected in mobile WebKit (Safari).
+       * @see https://github.com/sendbird/sendbird-uikit-react/pull/1108
+       */
+      if (isMobileIOS(navigator.userAgent)) {
+        if (ghostInputRef.current) ghostInputRef.current.focus();
+        requestAnimationFrame(() => textField.focus());
```
liamcho, line 405: > Using setTimeout or similar causes delays due to waiting for the event loop, resulting in the phenomenon of the virtual keyboard going up and down. Therefore, requestAnimationFrame, associated with the browser's cycle of drawing new frames, is used.

I didn't understand this sentence. requestAnimationFrame is using setTimeout with 0 second delay. Did you mean using with 0 second delay won't cause the up and down issue?
```
global.requestAnimationFrame = function (callback) {
  return setTimeout(callback, 0);
};
```

bang9: No, `requestAnimationFrame` is not using `setTimeout`

requestAnimationFrame is specifically designed to synchronize with the browser’s frame updates.
It queues the callback to execute right before the next repaint, ensuring that updates are smooth and in sync with the browser’s refresh cycle. This makes it much more suitable for animations and dynamic UI updates where timing and visual smoothness are critical.

setTimeout is an API used to delay the execution of a function. It executes the function once after a specified time in milliseconds set by the developer. Since this function runs when the event loop has available time after processing other tasks, the specified delay is the minimum wait time but the actual execution could be delayed further. As a result, it operates independently without synchronizing with other tasks in the browser.

liamcho: That's incredible this thing existed!

liamcho: In this case, what is triggering repaint? `ghostInputRef.current.focus();`?

liamcho (approved):
LGTM!

src/ui/MessageInput/index.tsx
bang9: Repaint happens when a browser redraws a web page to show the visual updates resulting from a UI change.

src/ui/MessageInput/index.tsx
bang9: Yes!

PR #1114, opened 2024-05-30
feat: support mov file type rendering only in Safari browser

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/SBISSUE-16031?focusedCommentId=270601 

Quoted a comment left by @jayden-lee-sb
>Not supporting ‘MOV’: This is not a bug since we display ‘MOV’ as a normal file message exactly described in the changelog(v3.13.2) → But we need to consider supporting ‘ThumbnailMessage’ only for some browser environments like Safari where MOV can be displayed without any problem

#### .mov file in Chrome browser
Screenshot 2024-05-29 at 5 24 43 PM

#### .mov file in Safari browser 
Screenshot 2024-05-29 at 5 25 43 PM

### DIFF
--- a/src/ui/FileViewer/types.ts
+++ b/src/ui/FileViewer/types.ts
@@ -1,40 +1,7 @@
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types
 
 // TODO: refactor this to -> as const pattern
-import { KeyboardEvent, MouseEvent } from 'react';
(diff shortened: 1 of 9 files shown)


src/ui/FileViewer/types.ts
AhyoungRyu, line 1: Removing these utils since we have the same ones in utils/index.ts. And it seems these have been added 3 years ago.

HoonBaek (approved):
LGTM!

PR #1123, opened 2024-06-04
[CLNP: 2977] fix: m4a format audio file playing in Safari

AhyoungRyu:
Fixes https://sendbird.atlassian.net/browse/CLNP-2997

#### Problem Statement

Voice audio files created on platforms other than the web, such as Android/iOS, are in M4A format. When attempting to convert these M4A files to `audio/mp3` or `audio/mpeg` format for playback on the web, Safari browser cannot find the correct codec and throws an error: `Unhandled Promise Rejection: NotSupportedError: The operation is not supported.`
This results in the file not being playable.

#### Solution

For Safari browser only, if the message file type is M4A, ensure it is played in this format. However, the file must be parsed as `audio/x-m4a` instead of `audio/m4a` to be correctly played in the Safari browser.

#### Test Cases Checked

| Browser  | .m4a file(created in iOS / Android UIKit) | .mp3 File(created in React UIKit) |
|----------|----------|----------|
| Safari   | ✅       | ✅       |
| Chrome   | ✅       | ✅       |
| Firefox  | ✅       | ✅       |

>  A testable account credential for .m4a can be found in the linked Jira ticket

### DIFF
--- a/src/hooks/VoicePlayer/index.tsx
+++ b/src/hooks/VoicePlayer/index.tsx
@@ -21,6 +21,7 @@ import {
   VOICE_PLAYER_ROOT_ID,
 } from '../../utils/consts';
 import useSendbirdStateContext from '../useSendbirdStateContext';
+import { getParsedMimeType } from './utils';
 
 // VoicePlayerProvider interface
 export interface VoicePlayerProps {
@@ -30,6 +31,7 @@ export interface VoicePlayerPlayProps {
   groupKey: string;
   audioFile?: File;
   audioFileUrl?: string;
+  audioFileMimeType?: string;
 }
 export interface VoicePlayerContext {
   play: (props: VoicePlayerPlayProps) => void;
@@ -91,6 +93,7 @@ export const VoicePlayerProvider = ({
     groupKey,
     audioFile,
     audioFileUrl = '',
+    audioFileMimeType = VOICE_MESSAGE_MIME_TYPE,
   }: VoicePlayerPlayProps): void => {
     if (groupKey !== currentGroupKey) {
       pause(currentGroupKey);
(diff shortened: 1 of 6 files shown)


HoonBaek (approved):
LGTM

src/utils/consts.ts
```diff
@@ -25,6 +25,7 @@ export const VOICE_PLAYER_AUDIO_ID = 'sendbird-global-audio-player-id';
 // voice message file
 export const VOICE_MESSAGE_FILE_NAME = 'Voice_message.mp3';
 export const VOICE_MESSAGE_MIME_TYPE = 'audio/mp3;sbu_type=voice';
+export const VOICE_MESSAGE_MIME_TYPE__M4A = 'audio/x-m4a;sbu_type=voice';
```
bang9, line 28:
```suggestion
export const VOICE_MESSAGE_MIME_TYPE__XM4A = 'audio/x-m4a;sbu_type=voice';
```

bang9: It would be good to explicitly use a x-m4a.

src/hooks/VoicePlayer/index.tsx
```diff
@@ -128,7 +131,7 @@ export const VoicePlayerProvider = ({
         .then((blob) => {
           const audioFile = new File([blob], VOICE_MESSAGE_FILE_NAME, {
             lastModified: new Date().getTime(),
-            type: VOICE_MESSAGE_MIME_TYPE,
+            type: getParsedMimeType(audioFileMimeType),
```
bang9, line 134: nit: When creating a file with a fetched blob, the file name contains an extension. Won't that be confusing later?

image

bang9 (approved):

src/hooks/VoicePlayer/index.tsx
AhyoungRyu: Oh yeah right. So I made it differentiate the name too. https://github.com/sendbird/sendbird-uikit-react/pull/1123/commits/04f1d6b333c6e35a2739540c3f3ed5b50853422a

bang9 (approved):

src/utils/consts.ts
AhyoungRyu: https://github.com/sendbird/sendbird-uikit-react/pull/1123/commits/04f1d6b333c6e35a2739540c3f3ed5b50853422a

PR #1126, opened 2024-06-07
feature: Add modal to the FileViewer components

HoonBaek:
[CLNP-3150](https://sendbird.atlassian.net/browse/CLNP-3150)

### Feature Request & Change Log
* Add a modal to the `FileViewer` components for using `onMounted` event handler

[CLNP-3150]: https://sendbird.atlassian.net/browse/CLNP-3150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ui/FileViewer/index.tsx
```diff
@@ -38,106 +39,108 @@ export const FileViewerComponent = (props: FileViewerComponentProps): ReactEleme
...
-      onKeyDown={onKeyDown}
-      // to focus
-      tabIndex={1}
-      ref={ref}
-    >
-      <div className="sendbird-fileviewer__header">
-        <div className="sendbird-fileviewer__header__left">
-          <div className="sendbird-fileviewer__header__left__avatar">
-            <Avatar height="32px" width="32px" src={profileUrl} />
+    <Modal>
```
HoonBaek, line 42: Just added `<Modal />`, no more change

AhyoungRyu (approved):

src/ui/FileViewer/index.tsx
liamcho: @HoonBaek In the ticket, I see no bug but what is the issue with not having Modal here?

HoonBaek: @liamcho 
The customer is using UIKit in a mobile environment, and on Android phones, there is a back button. To navigate back to the previous page using the back button, it is necessary to store the history for each page. We have supported this by allowing the customer to manage history through the `onMounted` event. The customer has requested that the same back button functionality, using history, be supported when navigating back from the `FileViewer` to the `Channel`.

liamcho: i got it. its for
```
  modal?: {
    onMounted?(params: { id: string; close(): void; }): void | (() => void);
  };
```
thanks!

liamcho: oh I understand the case now, thanks for explaining @HoonBaek :)

liamcho (approved):
LGTM!

PR #1128, opened 2024-06-10
[CLNP-3514] chore: remove onlight05 & ondark05

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-3514

It seems onlight-05 doesn't exist in other platforms and the product designer doesn't even know about the existence of this. So I'm removing and replacing it with onlight-03 instead. 

image
https://github.com/sendbird-graveyard/SendBird-UIKIT-JavaScript/tree/master#appendix


chrisallo (approved):
lgtm!

PR #1136, opened 2024-06-13
fix: dir is not being injected in Mobile layout

AhyoungRyu:


src/modules/CreateChannel/components/CreateChannelUI/create-channel-ui.scss
```diff
@@ -29,7 +29,6 @@
   @include mobile() {
     min-width: 288px;
     width: 100%;
-    padding-right: 0;
```
AhyoungRyu, line 32: The channel creation dialog layout won't be broken even without this. 

#### LTR
Screenshot 2024-06-13 at 4 36 26 PM

#### RTL
Screenshot 2024-06-13 at 4 36 22 PM

HoonBaek (approved):
LGTM

src/modules/App/hooks/useApplyTextDirection.ts
```diff
@@ -22,14 +23,15 @@ const ELEMENT_IDS = [
...
+  const { isMobile } = useMediaQueryContext();
   useEffect(() => {
     ELEMENT_IDS.forEach((id) => {
       const element = document.getElementById(id);
       if (element) {
         element.dir = direction;
       }
     });
-  }, [direction]);
+  }, [direction, isMobile]);
```
AhyoungRyu, line 34: `isMobile` is determined within `useMediaQueryContext`, and this needs to run separately of the `useApplyTextDirection` hook. This ensures that the internal logic of the `useApplyTextDirection` hook is executed only after the `isMobile` value has been properly set.

PR #1137, opened 2024-06-17
refactor: should disable in message input wrapper view

bang9:
## Changes
- Updated the logic to align with other platforms for consistency.
- Relocated the logic to the same section where other `disabled` conditions are checked.

ticket: [AC-2781]


src/modules/GroupChannel/context/utils.ts
```diff
@@ -35,3 +35,8 @@ export const isDisabledBecauseMuted = (groupChannel?: GroupChannel | null) => {
   if (!groupChannel) return false;
   return groupChannel.myMutedState === 'muted';
 };
+
+export const isDisabledBecauseSuggestedReplies = (channel: GroupChannel | null | undefined, enableSuggestedReplies: boolean) => {
+  if (!channel || !channel.lastMessage) return false;
+  return enableSuggestedReplies && !!channel.lastMessage.extendedMessagePayload?.disable_chat_input;
```
AhyoungRyu, line 41:
```suggestion
  return enableSuggestedReplies && !!channel?.lastMessage?.extendedMessagePayload?.disable_chat_input;
```
Actually we can get a same result even without the first line with the optional chainings.

AhyoungRyu (approved):
Left a minor comment but LGTM :)

src/modules/GroupChannel/context/utils.ts
bang9: -> https://github.com/sendbird/sendbird-uikit-react/pull/1137/commits/4fc8b62cc102f249795669021d115a01b17fb88c

PR #1142, opened 2024-06-19
fix: Export the UI MobileMenu

HoonBaek:
Export the MobileMenu

```tsx
import { MobileMenu, MobileContextMenu, MobileBottomSheet } from '@sendbird/uikit-react/ui/MobileMenu';
```

### DIFF
--- a/src/ui/MobileMenu/index.tsx
+++ b/src/ui/MobileMenu/index.tsx
@@ -13,4 +13,5 @@ const MobileMenu: React.FC<MobileBottomSheetProps> = (props: MobileBottomSheetPr
   );
 };
 
+export { MobileMenu, MobileContextMenu, MobileBottomSheet };
 export default MobileMenu;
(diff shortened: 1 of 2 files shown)


src/ui/MobileMenu/index.tsx
```diff
@@ -13,4 +13,5 @@ const MobileMenu: React.FC<MobileBottomSheetProps> = (props: MobileBottomSheetPr
   );
 };
 
+export { MobileMenu, MobileContextMenu, MobileBottomSheet };
 export default MobileMenu;
```
AhyoungRyu, line 17: Do we need to export in both ways?
 
```
export { MobileMenu };
export default MobileMenu;
```

HoonBaek: I think it's better for covering both case, we can deprecate one way in the future.
But until that time, we will keep both way.

AhyoungRyu (approved):

PR #1155, opened 2024-06-28
fix: Calculate the sendbird-app size for context menu position

HoonBaek:
[CLNP-4034](https://sendbird.atlassian.net/browse/CLNP-4034)

* Use the `APP_LAYOUT_ROOT` to get the area info of the UIKit
* Consider that area when calculating the correct `menu`s` position

[CLNP-4034]: https://sendbird.atlassian.net/browse/CLNP-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Sendbird.tsx
+++ b/src/lib/Sendbird.tsx
@@ -213,23 +213,23 @@ const SendbirdSDK = ({
     isUserIdUsedForNickname,
     isMobile,
   },
-  {
-    logger,
-    nickname,
-    profileUrl,
-    configureSession,
-    customApiHost,
-    customWebSocketHost,
-    sdkInitParams,
-    customExtensionParams,
-    sdk,
-    sdkDispatcher,
-    userDispatcher,
-    appInfoDispatcher,
-    initDashboardConfigs,
-    eventHandlers,
-    initializeMessageTemplatesInfo,
-  },
+    {
+      logger,
+      nickname,
+      profileUrl,
+      configureSession,
+      customApiHost,
+      customWebSocketHost,
+      sdkInitParams,
+      customExtensionParams,
+      sdk,
+      sdkDispatcher,
+      userDispatcher,
+      appInfoDispatcher,
+      initDashboardConfigs,
+      eventHandlers,
+      initializeMessageTemplatesInfo,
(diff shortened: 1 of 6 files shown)


src/lib/Sendbird.tsx
```diff
@@ -310,110 +311,112 @@ const SendbirdSDK = ({
...
-            threadReplySelectType: configs.groupChannel.channel.threadReplySelectType,
-            enableVoiceMessage: configs.groupChannel.channel.enableVoiceMessage,
-            enableDocument: configs.groupChannel.channel.input.enableDocument,
-            typingIndicatorTypes: configs.groupChannel.channel.typingIndicatorTypes,
-            enableFeedback: configs.groupChannel.channel.enableFeedback,
-            enableSuggestedReplies: configs.groupChannel.channel.enableSuggestedReplies,
-            showSuggestedRepliesFor: configs.groupChannel.channel.showSuggestedRepliesFor,
-            suggestedRepliesDirection: configs.groupChannel.channel.suggestedRepliesDirection,
-            enableMarkdownForUserMessage: configs.groupChannel.channel.enableMarkdownForUserMessage,
+    <div id="sendbird-provider-wrapper">
```
AhyoungRyu, line 314: 여긴 provider 의 역할만 하고  div 가 붙지 않는게 맞는거같아요. SendbirdProvider 를 쓰는 고객들도 갑자기 이 div 가 추가되면 div 가 가진 기본 스타일 설정들로 인해 원치않게 전체 스타일이 깨질 수도 있답니당

HoonBaek: How about now? I think we need to wrap all our components with this div tag to accurately determine the area covered by the UIKit, so that we can precisely calculate the position of the menu components.

AhyoungRyu: Hmm still, there might be an issue with adding an additional div element here since div has its default style settings like `display: block`.

AhyoungRyu (approved):

PR #1170, opened 2024-07-15
fix: Mark deprecation on Channel & ChannelList modules

HoonBaek:
- Add deprecation mark on `Channel`, `ChannelProvider`, `ChannelList`, and `ChannelListProvider`

### DIFF
--- a/src/modules/Channel/context/ChannelProvider.tsx
+++ b/src/modules/Channel/context/ChannelProvider.tsx
@@ -165,11 +165,11 @@ const ChannelContext = React.createContext<ChannelProviderInterface | null>(null
 /**
  * @deprecated This provider is deprecated and will be removed in future releases.
  * Please use the `GroupChannelProvider` from '@sendbird/uikit-react/GroupChannel/context' instead.
- * 
+ *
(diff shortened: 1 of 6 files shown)


HoonBaek:
Are there more places to add deprecation marks?

bang9:
How about adding a migration guide link instead of providing an example code?

src/modules/Channel/context/ChannelProvider.tsx
```diff
@@ -162,6 +162,12 @@ export interface ChannelProviderInterface extends ChannelContextProps, MessageSt
 
 const ChannelContext = React.createContext<ChannelProviderInterface | null>(null);
 
+/**
+ * @deprecated This provider is deprecated and will be removed in future releases.
```
bang9, line 166: I think `next major update` is better than `future release`. (`future release` is too general to determine a specific time.)

bang9:
Looks good to me! Let's wait for Irene's review.

bang9 (approved):

AhyoungRyu (approved):

PR #1175, opened 2024-07-18
feat: Add query params to the member list props

HoonBaek:
[CLNP-3679](https://sendbird.atlassian.net/browse/CLNP-3679)

## ChangeLog
* Added `memberListQueryParams` to the `MemberList`, `MembersModal`, `MutedMemberList`, and `MutedMembersModal`
* Added `bannedUserListQueryParams` to the `BannedUserList` and `BannedUsersModal`
* Added `operatorListQueryParams` to the `OperatorList` and `OperatorsModal`

[CLNP-3679]: https://sendbird.atlassian.net/browse/CLNP-3679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
+++ b/src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
@@ -72,6 +72,7 @@ export function MembersModal({
             <React.Fragment key={member.userId}>
               {
                 renderUserListItem({
+                  // NOTE: This `index` is used to display the current user's user item at the top when customizing externally.
                   index,
                   user: member,
                   channel,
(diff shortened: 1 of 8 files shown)


src/modules/ChannelSettings/components/ModerationPanel/MembersModal.tsx
```diff
@@ -65,10 +68,11 @@ export function MembersModal({
             },
           })}
         >
-          {members.map((member: Member) => (
+          {members.map((member: Member, index) => (
             <React.Fragment key={member.userId}>
               {
                 renderUserListItem({
+                  index,
```
AhyoungRyu, line 75: @HoonBaek Any reason we pass index here? It seems there's no usage of this index.

HoonBaek: I thought this will be useful when customer wants to add the current user item in Modals
Because in Modal, we can't divide only user list

AhyoungRyu (approved):

PR #1184, opened 2024-07-31
[CLNP-4596] feat: add forceLeftToRightMessageLayout to enable LTR message layout display in RTL mode

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-4596

This PR addresses the layout issues and CSS [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) improvements for right-to-left (RTL) text direction in the Sendbird UIKit React components. 

The background for adding the `forceLeftToRightMessageLayout` flag is that some of our users in the middle-east want to set `htmlTextDirection='rtl'` while keeping the message layout in left-to-right (LTR) format (outgoing messages on the right, incoming messages on the left).

### Todo
- [x] The postcssRtl plugin options were modified.
- [x] Layout adjustments were made to ensure proper alignment and behavior.

### DIFF
--- a/postcssRtlOptions.mjs
+++ b/postcssRtlOptions.mjs
@@ -15,3 +15,50 @@ export default {
     return `${prefix} ${selector}`;
   },
 }
+
+// Why we're doing this:
+// Let's say a root div element has `dir="rtl"` (with `htmlTextDirection={'rtl'}` prop setting).
+
+/*
+<div class="sendbird-root" dir="rtl">
+  ...
+  // Message list
+  <div class="sendbird-conversation__messages" dir="ltr">
+     <div class="emoji-container">
+       <span class="emoji-inner">😀</span>
+     </div>
+  </div>
+</div>
+*/
+
+// Even though the message list element has dir="ltr" attribute,
+// some children elements of the message list would have their styles overridden by the root one with the CSS settings below.
+// Why? postcss-rtlcss plugin generates the CSS settings in order of rtl -> ltr.
+
+/*
+[dir="rtl"] .sendbird-message-content .emoji-container .emoji-inner {
+    right: -84px; // Specificity (0.6.0)
+}
+[dir="ltr"] .sendbird-message-content .emoji-container .emoji-inner {
+    left: -84px; // Specificity (0.6.0)
+}
+*/
+
+// If both CSS settings have the same specificity, the one generated first (rtl) will be applied,
+// which is not the desired result since we want the `.emoji-inner` element to have the ltr setting.
+
+// To increase the specificity of the ltr setting,
+// we can directly connect the classname of the element which has `dir='ltr'` to the children's selector in this way:
+
+/*
+.sendbird-conversation__messages[dir="ltr"] .sendbird-message-content .emoji-container .emoji-inner {
+  left: -84px; // Specificity (0.7.0), will be applied
+}
+
+[dir="rtl"] .sendbird-message-content .emoji-container .emoji-inner {
+    right: -84px; // Specificity (0.6.0), will be ignored
+}
+[dir="ltr"] .sendbird-message-content .emoji-container .emoji-inner {
+    left: -84px; // Specificity (0.6.0), will be ignored
+}
+*/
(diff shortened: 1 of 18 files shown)


AhyoungRyu:
For whom wants to test this 
- RTL but only the message layout keeps LTR : http://localhost:5173/?htmlTextDirection=rtl&forceLeftToRightMessageLayout=true
ltr

- Normal RTL: https://deploy-preview-1184--sendbird-uikit-react.netlify.app?htmlTextDirection=rtl
rtl

postcssRtlOptions.mjs
```diff
@@ -0,0 +1,64 @@
...
+      'sendbird-thread-list-item-content',
+      'sendbird-parent-message-info'
+    ].some(cls => selector.includes(cls))) {
+      return `.sendbird-conversation__messages${prefix} ${selector}`;
+    }
+    return `${prefix} ${selector}`;
+  },
+}
+
+// Why we're doing this:
```
AhyoungRyu, line 19: Added more explanation for who might wonder

src/hooks/useHTMLTextDirection.tsx
```diff
@@ -15,4 +15,18 @@ const useHTMLTextDirection = (direction: HTMLTextDirection) => {
   }, [direction]);
 };
 
+export const useMessageLayoutDirection = (direction: HTMLTextDirection, forceLeftToRightMessageLayout: boolean, loading: boolean) => {
+  useEffect(() => {
+    if (loading) return;
+    const messageListElements = document.getElementsByClassName('sendbird-conversation__messages');
```
HoonBaek, line 21: I think it's good to cover this scope with `try catch`

postcssRtlOptions.mjs
```diff
@@ -0,0 +1,64 @@
...
+  prefixSelectorTransformer: (prefix, selector) => {
+    // To increase specificity https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity 
+    // for certain classnames within .sendbird-conversation__messages context.
+    // This only applies when the root dir is set as 'rtl' but forceLeftToRightMessageLayout is true.
+    if ([
+      // Normal message list
+      'sendbird-message-content',
+      // Thread message list
+      'sendbird-thread-list-item-content',
+      'sendbird-parent-message-info'
```
HoonBaek, line 11: How about making this Array as const? like this
```tsx
const CLASS_NAMES_TO_CHECK = [
  // Normal message list
  'sendbird-message-content',
  // Thread message list
  'sendbird-thread-list-item-content',
  'sendbird-parent-message-info'
];
```

```tsx
CLASS_NAMES_TO_CHECK.some(...)
```

src/modules/Thread/components/ThreadUI/index.tsx
```diff
@@ -148,7 +148,7 @@ const ThreadUI: React.FC<ThreadUIProps> = ({
         )
       }
       <div
-        className="sendbird-thread-ui--scroll"
+        className={'sendbird-thread-ui--scroll sendbird-conversation__messages'}
```
HoonBaek, line 151: let's use the `classnames` function
```tsx
className={classnames('sendbird-thread-ui--scroll', 'sendbird-conversation__messages')}
```

HoonBaek (approved):
LGTM!

src/hooks/useHTMLTextDirection.tsx
AhyoungRyu: Oh if there's no element to target, this will return an empty array if you're worry about this :)

postcssRtlOptions.mjs
AhyoungRyu: Yeah addresses in https://github.com/sendbird/sendbird-uikit-react/pull/1184/commits/52397fd40e6d71fb54851ed276da8c1d950cafa3

src/modules/Thread/components/ThreadUI/index.tsx
AhyoungRyu: This one too https://github.com/sendbird/sendbird-uikit-react/pull/1184/commits/52397fd40e6d71fb54851ed276da8c1d950cafa3

PR #1213, opened 2024-09-04
hotfix: Channel UI's scroll not work after sending new message

git-babel:


src/modules/GroupChannel/context/hooks/useMessageActions.ts
```diff
@@ -75,13 +75,22 @@ export function useMessageActions(params: Params): MessageActions {
     [replyType, quoteMessage],
   );
 
+  // This is a hack for the hotfix of following issue
+  // https://sendbird.atlassian.net/browse/SBISSUE-17029
+  const asyncScrollToBottom = useCallback(
```
AhyoungRyu, line 80:
```suggestion
  const throttledScrollToBottom = useThrottleCallback(scrollToBottom, 0);
```
Could you check if this works same as expected? Just found a util function we already have.

git-babel: The issue is resolved with the following code. trailing option is mandatory cause it makes callback to async.
```
const throttledScrollToBottom = useThrottleCallback(scrollToBottom, 100, { trailing: true });
```
But in my opinion, it is deviated usage from the original purpose of `useThrottleCallback`. As it says, it was intended to be used for throttling logics, which repeat very rapidly, like scroll or resize event. So in robust view, I think using that utility function is not proper. I want to encapsulate some hack code in its own file, rather than connect to an existing logic.

AhyoungRyu: OK cool. Thanks for checking on it.

AhyoungRyu (approved):

PR #1228, opened 2024-10-08
[CLNP-5043] Add unit tests for improved MessageSearch module

AhyoungRyu:
https://sendbird.atlassian.net/browse/CLNP-5043 
Two things are handled based on what I mentioned in https://sendbird.atlassian.net/wiki/spaces/UIKitreact/pages/2511765635/UIKit+React+new+State+Management+Method+Proposal#4.1-Unit-Test 
- [x] Added unit tests for `useMessageSearch` hook and new `MessageSearchProvider` 
- [x] Added integration tests for `MessageSearchUI` component

So the MessageSearch module test coverage has been changed 
**from** 
File --------------------------------------------------| % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
Screenshot 2024-10-08 at 2 36 55 PM
**to** 
after
note that it used to be like 0%, but now the test coverage of the newly added files is almost 100%; green 🟩. 

### Checklist


src/modules/MessageSearch/context/__test__/useMessageSearch.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('updates state when setQueryInvalid is called', () => {
+    const { result } = renderHook(() => useMessageSearch(), { wrapper });
+
+    act(() => {
+      result.current.actions.setQueryInvalid();
+    });
+
+    expect(result.current.state.isInvalid).toBe(true);
```
chrisallo, line 90: While I was reading the code and the test, it was a bit confusing for me what `isInvalid` means because both a channel and a query could go invalid. Do you mind changing the name to `isQueryInvalid` for clarity?

src/modules/MessageSearch/context/__test__/MessageSearchProvider.spec.tsx
```diff
@@ -0,0 +1,157 @@
...
+      </MessageSearchProvider>
+    );
+
+    const { result } = renderHook(() => useMessageSearch(), { wrapper });
+
+    await act(async () => {
+      result.current.actions.setCurrentChannel({ url: 'test-channel' } as any);
+      await waitFor(() => {
+        expect(result.current.state.currentChannel).toEqual({ url: 'test-channel' });
+        expect(result.current.state.initialized).toBe(true);
```
chrisallo, line 106: It makes more sense to me to make sure the other states such as `loading` and `isInvalid` unchanged. It would help finding some regression before the thing could go into trouble imho.

src/modules/MessageSearch/context/__test__/useMessageSearch.spec.tsx
AhyoungRyu: Yeah I got your point and don't mind at all. Updated the prop name in here https://github.com/sendbird/sendbird-uikit-react/pull/1228/commits/35702bac83c4f072e4cbcc6faed8f6d6211e2ed9

src/modules/MessageSearch/context/__test__/MessageSearchProvider.spec.tsx
AhyoungRyu: Thanks. So I updated other tests as well https://github.com/sendbird/sendbird-uikit-react/pull/1228/commits/b6849336baa909aa9164660b492eea369d9644be

chrisallo (approved):
lgtm!

PR #1231, opened 2024-10-11
refactor: GroupChannelList migration

git-babel:
### Changelogs
This is a part of state management migration. This PR migrate 'GroupChannelListProvider' and related files to the new style.
* Created `useGroupChannelList()` hook. It'll replace the previous `useGroupChannelContext()` hook.

### DIFF
--- a/src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
+++ b/src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 
 import type { User } from '@sendbird/chat';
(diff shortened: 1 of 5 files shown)


git-babel:
There is no action for the `useGroupChannelList()` hook for there is no external state change action for this provider.

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
```diff
@@ -120,33 +154,71 @@ export const GroupChannelListProvider = (props: GroupChannelListProviderProps) =
     },
   });
 
+  useEffect(() => {
```
git-babel, line 157: I thought the props of provider and internal state handled inside the provider should be here. Is it right?

HoonBaek (approved):

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
AhyoungRyu: @git-babel Yeah it is correct

AhyoungRyu (approved):

PR #1233, opened 2024-10-18
refactor: Add unit tests for new GroupChannelListProvider

git-babel:
This PR contains the unit tests for the recent refactoring of `GroupChannelListProvider`.
* add test about the new provider and its state management logic itself.
* add the integration test with `GroupChannelListUI` component.


git-babel:
I have a question. We have some lines like below in `GroupChannelListProvider`
```javascript
  const channelListDataSource = useGroupChannelList(sdk, {
    collectionCreator: getCollectionCreator(sdk, channelListQueryParams),
    markAsDelivered: (channels) => channels.forEach(scheduler.push),
    onChannelsDeleted: (channelUrls) => {
      channelUrls.forEach((url) => {
        if (url === selectedChannelUrl) onChannelSelect(null);
      });
    },
  });
  const { refreshing, initialized, groupChannels, refresh, loadMore } = channelListDataSource;
```
that `useGroupChannelList` comes from `@sendbird/uikit-tools` package. I'm not sure that we have to include those state from `channelListDataSource` in our unit test.
It seems like that the `sendbird/uikit-tools` already handle the state updates. And it was hard to call some appropriate action and observe the state change cause those logic are inside of `@sendbird/uikit-tools`. For now, those states from `channelListDataSource` is excluded in this PR's test cases. Should I add them?

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
```diff
@@ -66,12 +65,12 @@ const initialState: GroupChannelListState = {
...
-  onThemeChange: noop,
-  onCreateChannelClick: noop,
-  onBeforeCreateChannel: null,
-  onUserProfileUpdated: noop,
+  onChannelSelect: undefined,
+  onChannelCreated: undefined,
+  onThemeChange: undefined,
+  onCreateChannelClick: undefined,
+  onBeforeCreateChannel: undefined,
+  onUserProfileUpdated: undefined,
```
HoonBaek, line 73: We don't handle the undefined functions on the other files, like optional chaining.

We should confirm if there's no Function undefined error, during run-time.
or use `noop` again

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
```diff
@@ -191,7 +193,7 @@ export const GroupChannelListManager: React.FC<GroupChannelListProviderProps> =
     typingChannelUrls,
     refreshing,
     initialized,
-    groupChannels,
+    // groupChannels,
```
HoonBaek, line 196: Is this a typo?

AhyoungRyu:
Just checked the test coverage and it seems the newly created `useGroupChannelList` file doesn't have corresponding test file. Could you add one too? Also, I hope the coverage of  `GroupChannelListProvider` can be green not yellow. 
- Before this change 
Screenshot 2024-10-21 at 2 57 41 PM

- After this change 
Screenshot 2024-10-21 at 2 59 56 PM


You can also get some hint from here https://github.com/sendbird/sendbird-uikit-react/pull/1234#issuecomment-2425652104 about the way you increase the coverage. It can be checked with `yarn test:coverage` command on the root of this project.

AhyoungRyu:
>For now, those states from channelListDataSource is excluded in this PR's test cases. Should I add them?

Don't need to add for it since the logic is from mostly outside of the main logic we want to test.

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
git-babel: This is updated like the comment below by Irene.

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
git-babel: Fixed.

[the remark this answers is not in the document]
git-babel: Applied.

src/modules/GroupChannelList/context/__tests__/GroupChannelListProvider.spec.js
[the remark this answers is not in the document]
git-babel: I changed it to '.tsx'

src/modules/GroupChannelList/context/__tests__/useGroupChannelList.spec.js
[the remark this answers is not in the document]
git-babel: I changed it to '.tsx'

git-babel:
I added and updated the tests for new provider and hook. The result is like below.
스크린샷 2024-10-23 오전 11 07 19
The unvisited lines in `GroupChannelListProvider` are the callback functions to be called in `@sendbird/uikit-tools`. The actual implementation of functions from `@sendbird/uikit-tools` is shadowed by the mocks in test file, because originally it works only when the valid data is given. So it is hard to cover those lines now.

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
```diff
@@ -87,27 +87,26 @@ const useGroupChannelListStore = () => {
...
+  selectedChannelUrl = '',
 
-    disableAutoSelect = false,
-    allowProfileEdit,
-    isTypingIndicatorEnabled,
-    isMessageReceiptStatusEnabled,
+  disableAutoSelect = false,
+  allowProfileEdit = false,
+  isTypingIndicatorEnabled = false,
+  isMessageReceiptStatusEnabled = false,
```
AhyoungRyu, line 97:
```suggestion
  allowProfileEdit,
  isTypingIndicatorEnabled,
  isMessageReceiptStatusEnabled,
```
We'd better to keep the original status of this prop as-is since there might be a chance that some behavior changes due to this default value setting.

AhyoungRyu (approved):

HoonBaek (approved):
👍

src/modules/GroupChannelList/context/GroupChannelListProvider.tsx
git-babel: Applied.

PR #1234, opened 2024-10-18
refactor: Add tests for ChannelSettings migration

HoonBaek:
fix: Prevent destroy error by adding `AbortController` in `useSetChannel`
- Added `AbortController` to cancel async operations when the component unmounts.
- Ensured that state updates are skipped if the operation is aborted.
- Prevented potential memory leaks and the `'destroy is not a function'` error.
- Updated `useEffect` cleanup to properly handle pending async calls.

feat: Add tests for ChannelSettings migration

### DIFF
--- a/src/modules/ChannelSettings/__test__/ChannelSettingsProvider.spec.tsx
+++ b/src/modules/ChannelSettings/__test__/ChannelSettingsProvider.spec.tsx
@@ -1,0 +1,93 @@
+import React from 'react';
+import { renderHook, act } from '@testing-library/react-hooks';
+import { ChannelSettingsProvider, useChannelSettingsContext } from '../context/ChannelSettingsProvider';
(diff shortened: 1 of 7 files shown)


AhyoungRyu:
Leaving the test coverage for the ChannelSettings module before and after this change
- Before 
Screenshot 2024-10-21 at 2 44 47 PM

- After 
Screenshot 2024-10-21 at 2 43 17 PM

HoonBaek:
The hook which calls `setLoading` and `setInvalid` is the `useSetChannel` hook.
And I made a separate test for it, so, should I remove the two tests?
@AhyoungRyu

AhyoungRyu:
@HoonBaek Yeah I think we can remove them if that's the case.

HoonBaek:
image
Test coverage hasn't been changed after removing the two tests, let me remove them!

AhyoungRyu (approved):
👍

PR #1239, opened 2024-10-24
fix: Update suggested replies logic

liamcho:
Fixes [AC-4063](https://sendbird.atlassian.net/browse/AC-4063)

### 요약
일반 봇이 메세지를 보내면 channel.lastMessage 가 업데이트 안된답니다. 그래서 channel.lastMessage 에 의존하는 코드를 수정했습니다.

### Changelogs
- improved stability

[AC-4063]: https://sendbird.atlassian.net/browse/AC-4063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Channel/components/Message/index.tsx
```diff
@@ -51,7 +52,7 @@ const Message = (props: MessageProps) => {
         && (
           config?.groupChannel?.showSuggestedRepliesFor === 'all_messages'
             ? true
-            : message.messageId === currentGroupChannel?.lastMessage?.messageId
+            : message.messageId === allMessages[allMessages.length - 1].messageId
```
HoonBaek, line 55: It will be good if we comment why we don't use `lastMessage` here

liamcho: showSuggestedRepliesFor can have two values: all_messages, and last_message_only. if last_message_only, we render only if message is the last one.

HoonBaek: I'm sorry, I was talking about leaving a comment on the line 😅

src/modules/GroupChannel/components/Message/index.tsx
```diff
@@ -45,8 +45,9 @@ export const Message = (props: MessageProps): React.ReactElement => {
 
     const lastMessageInView = messages[messages.length - 1];
```
HoonBaek, line 46: good to add here too

HoonBaek (approved):

PR #1254, opened 2024-11-20
fix: improve error handling in message handlers

AhyoungRyu:
Fixes some of comments in [SBISSUE-17017](https://sendbird.atlassian.net/browse/SBISSUE-17017?focusedCommentId=298138)

## Problem
Users currently need to return an empty object from `onBeforeSendFileMessage` due to type constraints. Additionally, error handling for message sending failures needs improvement to ensure `eventHandlers.message` callbacks (onSendMessageFailed, onUpdateMessageFailed, onFileUploadFailed) are properly triggered when errors occur within their corresponding `onBefore` handlers (`onBeforeSendUserMessage`, `onBeforeSendFileMessage`, etc).

## Solution
- Allow void return type in `onBeforeXXX` handler
- Add proper error handling via `eventHandlers.message` for:
  - onSendMessageFailed
  - onUpdateMessageFailed
  - onFileUploadFailed

### Checklist

[SBISSUE-17017]: https://sendbird.atlassian.net/browse/SBISSUE-17017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "@svgr/rollup": "^8.1.0",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
+    "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^14.4.3",
     "@types/node": "^22.7.2",
     "@typescript-eslint/eslint-plugin": "^6.17.0",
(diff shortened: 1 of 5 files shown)


HoonBaek (approved):
LGTM!

bang9:
@HoonBaek @AhyoungRyu The `onBeforeHandler` is essentially something the user injects, so I don't think we need to handle errors for them—they should be able to handle that themselves. 

What the customer seems to want is to handle failures when sending messages using SDK.
Shouldn't it be handled like this?

```javascript
try {
  const params = await onBeforeSendUserMessage(...);
  return await sendUserMessage(params); // -> here
} catch (error) {
  // Error: Receiver is blocked
  eventHandlers.message.onSendMessageFailed(error);
  throw error;
}
```

PR #1259, opened 2024-11-29
fix: Make tests passible for SendbirdProvider migration

HoonBaek:
* Make it passible the tests affected by SendbirdProvider migration

### DIFF
--- a/src/hooks/VoicePlayer/index.tsx
+++ b/src/hooks/VoicePlayer/index.tsx
@@ -20,7 +20,7 @@ import {
   VOICE_PLAYER_ROOT_ID,
 } from '../../utils/consts';
 import { getParsedVoiceAudioFileInfo } from './utils';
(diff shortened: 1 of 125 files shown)


src/hooks/VoicePlayer/index.tsx
```diff
@@ -20,7 +20,7 @@ import {
   VOICE_PLAYER_ROOT_ID,
 } from '../../utils/consts';
 import { getParsedVoiceAudioFileInfo } from './utils';
-import useSendbird from '../../lib/Sendbird/context/hooks/useSendbird';
+import { useSendbird } from '../../lib/Sendbird/context/hooks/useSendbird';
```
AhyoungRyu, line 23: @HoonBaek  Why don't we just make a default export from useSendbird?

AhyoungRyu (approved):

PR #1268, opened 2024-12-02
[CLNP-5914] Fixed a GroupChannel UI error when the `Open in Channel` action is triggered in a different channel.

git-babel:
### Changelog
* Fixed a GroupChannel UI error when the `Open in Channel` action is triggered in a different channel.
  * The `Open in Channel` implementation invokes both `setCurrentChannel` and `setStartingPoint`. The `setCurrentChannel` function triggers asynchronous side effects that update `channel`, `messagesDataSource`, and `startingPoint`. If `setStartingPoint` is invoked before these updates are completed, it can result in the channel not being updated correctly or the starting point being improperly set.


AhyoungRyu:
@git-babel Could you elaborate more about what the fix is in the PR description or title?

AhyoungRyu (approved):

PR #1276, opened 2024-12-05
fix(channel): show members as titles by default for AI Chatbot Widget Channel

aldenquimby:
I know you cannot accept PRs from External Contributions.

I just filed a support ticket for this, and am creating a PR to show how I think it could be addressed

Here is what things look like before the PR:
image

And after:
image

The idea is that for AI Chatbot generated channels, we should display the name of the chatbot, rather than the default generated `AI Chatbot Widget Channel`. And only display the channel name if someone as modified it to something custom


src/modules/GroupChannelList/components/GroupChannelListItem/utils.ts
```diff
@@ -10,7 +10,7 @@ export const getChannelTitle = (channel?: GroupChannel, currentUserId?: string,
   if (!channel?.name && !channel?.members) {
     return stringSet.NO_TITLE;
   }
-  if (channel?.name && channel.name !== 'Group Channel') {
+  if (channel?.name && channel.name !== 'Group Channel' && channel.name !== 'AI Chatbot Widget Channel') {
```
aldenquimby, line 13: Here is the primary fix. The rest of the PR is getting all consumers to use a single getChannelTitle helper function

src/modules/MessageSearch/components/MessageSearchUI/index.tsx
```diff
@@ -108,7 +98,7 @@ export const MessageSearchUI: React.FC<MessageSearchUIProps> = ({
       <div className="sendbird-message-search">
         <PlaceHolder
           type={PlaceHolderTypes.SEARCH_IN}
-          searchInString={getChannelName()}
+          searchInString={getChannelTitle(currentChannel, '', stringSet)}
```
aldenquimby, line 101: I wasn't sure if it'd be safe to add this hook to this component:
```ts
  const state = useSendbirdStateContext();
  const userId = state?.config?.userId;
```

(passing `userId = ''` should be safe if that hook is no good... it'd mean no behavior change here - the prior `getChannelName` function was already not incorporating currentUserId when joining members)

git-babel:
@aldenquimby Thank you for your PR. But as you said, we don't accept the external contribution. Instead, we made the fix for this issue. The fix will be included in next `3.15.0` release.

PR #1283, opened 2024-12-10
[CLNP-6003] test: Add tests for hooks in `src/hooks`

git-babel:
### Changelog
* Added tests for hooks in `src/hooks`

#### before
스크린샷 2024-12-10 오후 4 09 19

#### after
스크린샷 2024-12-10 오후 4 08 14


src/hooks/__tests__/useAppendDomNode.spec.tsx
AhyoungRyu, line 1: It seems nowhere is using this hook. So I think we can remove the file 😅

git-babel: I removed the file.

AhyoungRyu (approved):

PR #1285, opened 2024-12-11
Fix migration test failure

git-babel:
### Description
Fixed the failures of migration tests added by #1279


src/modules/MessageSearch/context/__tests__/MessageSearch.migration.spec.tsx
```diff
@@ -59,14 +59,11 @@ describe('MessageSearch Migration Compatibility Tests', () => {
...
       'requestString',
       'retryCount',
-      'setRetryCount',
       'selectedMessageId',
       'setSelectedMessageId',
-      'messageSearchDispatcher',
       'scrollRef',
       'allMessages',
       'loading',
-      'isInvalid',
```
AhyoungRyu, line 69: Let me check these why they're not in the return value of new useMessageSearchContext

AhyoungRyu: Put `setRetryCount` and `isInvalid` in https://github.com/sendbird/sendbird-uikit-react/pull/1285/commits/e778ac1a2bb06cbe75faa509861889d65f0c2dc7 but didn't keep `messageSearchDispatcher` since we dropped the legacy dispatcher logic. I'll mention in the CHANGELOG as well when we release the state migration.

AhyoungRyu (approved):

PR #1286, opened 2024-12-12
release: v3.15.10

HoonBaek:
### Features:
- Added `scrollRef` in GroupChannelList context

### Fixes:
- Fixed broken UI
  * MessageInput height becomes short when it's disabled.
  * Empty UserListItem menu appears on the ChannelSettings member list of normal channel member who is not an operator of the channel.
- Displayed members' name instead of default AI chatbot channel name, like the `Group Channel` does.
- Fixed an issue where editing a text parent message in a channel did not update the corresponding parent message in the Thread area in real-time.
- Fixed a GroupChannel UI error when the `Open in Channel` action is triggered in a different channel.
  * The `Open in Channel` implementation invokes both `setCurrentChannel` and `setStartingPoint`. The `setCurrentChannel` function triggers asynchronous side effects that update `channel`, `messagesDataSource`, and `startingPoint`. If `setStartingPoint` is invoked before these updates are completed, it can result in the channel not being updated correctly or the starting point being improperly set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@
   * Empty UserListItem menu appears on the ChannelSettings member list of normal channel member who is not an operator of the channel.
 - Fixed an issue where editing a text parent message in a channel did not update the corresponding parent message in the Thread area in real-time.
 - Fixed a GroupChannel UI error when the `Open in Channel` action is triggered in a different channel.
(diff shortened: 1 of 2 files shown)


sendbird-sdk-deployment:
[Creating Ticket] 🔖 Done https://sendbird.atlassian.net/browse/SDKRLSD-1497

AhyoungRyu (approved):

PR #1287, opened 2024-12-12
[CLNP-6032]Display members' name instead of default AI chatbot channel name

git-babel:
https://sendbird.atlassian.net/browse/CLNP-6032

### Changelog
* Display members' name instead of default AI chatbot channel name, like the `Group Channel` does.

### DIFF
--- a/src/modules/ChannelSettings/components/ChannelProfile/index.tsx
+++ b/src/modules/ChannelSettings/components/ChannelProfile/index.tsx
@@ -28,10 +28,10 @@ const ChannelProfile: React.FC = () => {
   const channel = channelSettingStore?.channel;
 
   const channelName = useMemo(() => {
-    if (channel?.name && channel.name !== 'Group Channel') {
+    if (channel?.name && channel.name !== 'Group Channel' && channel.name !== 'AI Chatbot Widget Channel') {
       return channel.name;
     }
-    if (channel?.name === 'Group Channel' || !channel?.name) {
+    if (channel?.name === 'Group Channel' || channel?.name === 'AI Chatbot Widget Channel' || !channel?.name) {
       return (channel?.members || []).map((member) => member.nickname || stringSet.NO_NAME).join(', ');
     }
 
--- a/src/modules/ChannelSettings/components/ChannelProfile/index.tsx
+++ b/src/modules/ChannelSettings/components/ChannelProfile/index.tsx
@@ -13,6 +13,7 @@ import Label, {
 } from '../../../../ui/Label';
 
 import EditDetailsModal from '../EditDetailsModal';
+import { isDefaultChannelName } from '../../../../utils';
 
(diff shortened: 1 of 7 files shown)


src/modules/ChannelSettings/components/ChannelProfile/index.tsx
```diff
@@ -28,10 +28,10 @@ const ChannelProfile: React.FC = () => {
   const channel = channelSettingStore?.channel;
 
   const channelName = useMemo(() => {
-    if (channel?.name && channel.name !== 'Group Channel') {
+    if (channel?.name && channel.name !== 'Group Channel' && channel.name !== 'AI Chatbot Widget Channel') {
```
AhyoungRyu, line 31: Let's manage this `AI Chatbot Widget Channel` string separately in somewhere else so we don't need to compare the string directly every time.

AhyoungRyu: Also, shall we create a separate util function for getting this condition ? 
```
if (channel?.name && channel.name !== 'Group Channel' && channel.name !== 'AI Chatbot Widget Channel') {
```

git-babel: `isDefaultChannelName()` is added in `src/utils/index.ts`. Also, `DEFAULT_GROUP_CHANNEL_NAME` and `DEFAULT_AI_CHATBOT_CHANNEL_NAME` are introduced too. But those consts are only used in `isDefaultChannelName()` for now.

src/modules/ChannelSettings/components/ChannelProfile/index.tsx
```diff
@@ -28,10 +29,10 @@ const ChannelProfile: React.FC = () => {
   const channel = channelSettingStore?.channel;
 
   const channelName = useMemo(() => {
-    if (channel?.name && channel.name !== 'Group Channel') {
+    if (!isDefaultChannelName(channel)) {
       return channel.name;
     }
-    if (channel?.name === 'Group Channel' || !channel?.name) {
+    if (isDefaultChannelName(channel)) {
```
AhyoungRyu, line 35: It can be just 
```
if (isDefaultChannelName(channel)} { return ... }
else { return channel.name }
```

AhyoungRyu (commented):
Same comments for other files.

src/modules/ChannelSettings/components/ChannelProfile/index.tsx
git-babel: I Applied it. And additionally, I aligned the logic for the `stringSet.NO_TITLE` like below.
```
if (!channel?.name && !channel?.members) return stringSet.NO_TITLE;
```

AhyoungRyu (approved):

PR #1288, opened 2024-12-16
[CLNP-6022] Memoize useGroupChannel actions to prevent unnecessary re-rendering

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/CLNP-6022?focusedCommentId=301263

### Key changes 
Memoized action handlers(+ scroll related functions as well) in useGroupChannel to reduced unnecessary re-rendering.

### DIFF
--- a/src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
+++ b/src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
@@ -1,4 +1,4 @@
-import React, { DependencyList, forwardRef, UIEventHandler, useLayoutEffect, useRef } from 'react';
+import React, { DependencyList, forwardRef, UIEventHandler, useCallback, useLayoutEffect, useRef } from 'react';
 import type { BaseMessage } from '@sendbird/chat/message';
 import { isAboutSame } from '../../../Channel/context/utils';
 import { SCROLL_BUFFER } from '../../../../utils/consts';
@@ -61,7 +61,7 @@ export const InfiniteList = forwardRef((props: Props, listRef: React.RefObject<H
     }
   }, [listRef.current, messages.length]);
 
-  const handleScroll: UIEventHandler<HTMLDivElement> = async () => {
+  const handleScroll: UIEventHandler<HTMLDivElement> = useCallback(async () => {
     if (!listRef.current) return;
     const list = listRef.current;
 
@@ -87,7 +87,7 @@ export const InfiniteList = forwardRef((props: Props, listRef: React.RefObject<H
     } else {
       direction.current = undefined;
     }
-  };
+  }, [listRef.current, messages.length]);
 
   return (
     <div className="sendbird-conversation__scroll-container">
--- a/src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
+++ b/src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
@@ -87,7 +87,7 @@ export const InfiniteList = forwardRef((props: Props, listRef: React.RefObject<H
     } else {
       direction.current = undefined;
     }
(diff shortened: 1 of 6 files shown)


src/modules/GroupChannel/context/hooks/useGroupChannel.ts
AhyoungRyu, line 1: Just wrapped all handlers with useCallback one by one. No logic changes.

src/utils/storeManager.ts
```diff
@@ -7,7 +9,16 @@ export type Store<T> = {
...
+    if (typeof prevState[key as keyof T] === 'function' && typeof value === 'function') {
+      /**
+       * Function is not considered as state change. Why?
+       * Because function is not a value, it's a reference.
+       * If we consider non-memoized function as state change,
+       * it will always be true and cause unnecessary re-renders.
+       */
+      return false;
+    }
+    return !isEqual(prevState[key as keyof T], value);
```
AhyoungRyu, line 21: Changed the way of comparing prev <-> next state with `isEqual` for deeper comparison[ like useDeepCompareEffect](https://github.com/sendbird/sendbird-uikit-react/blob/feat/state-mgmt-migration-1/src/hooks/useDeepCompareEffect.ts#L7)

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
```diff
@@ -87,7 +87,7 @@ export const InfiniteList = forwardRef((props: Props, listRef: React.RefObject<H
     } else {
       direction.current = undefined;
     }
-  };
+  }, [listRef.current, messages.length]);
```
HoonBaek, line 90: How about using `listRef.current?.scrollHeight, listRef.current?.scrollTop, listRef.current?.clientHeight` instead of `listRef.current`?

src/modules/GroupChannel/context/GroupChannelProvider.tsx
```diff
@@ -143,8 +144,8 @@ const GroupChannelManager :React.FC<React.PropsWithChildren<GroupChannelProvider
         channels.forEach((it) => markAsReadScheduler.push(it));
       }
     },
-    onMessagesReceived: () => {
-      if (isScrollBottomReached && isContextMenuClosed()) {
+    onMessagesReceived: (messages) => {
+      if (isScrollBottomReached && isContextMenuClosed() && messages.length !== state.messages.length) {
```
HoonBaek, line 148: 이런 케이스가 발생할 수 있나요?

HoonBaek (approved):

src/modules/GroupChannel/components/MessageList/InfiniteList.tsx
AhyoungRyu: Oh actually we don't even need to add `ref.current` in here cause it's a mutable object. So I removed it. https://github.com/sendbird/sendbird-uikit-react/pull/1288/commits/f9168a03b7c020a0dd114cac2e3a512a25534b1a#diff-5304e2564a5f57c7d83cf050597f6f595beef659c04a8156d25b1e203bd2281bL90

src/modules/GroupChannel/context/GroupChannelProvider.tsx
AhyoungRyu: Yeah ideally it shouldn't happen, but it's happening. I think we could handle this kind of optimization in the useGroupChannelMessages hook. Until then, this kind of handling can do some work. I added some comments for whom might wonder. https://github.com/sendbird/sendbird-uikit-react/pull/1288/commits/c1f6babcf0dca78904a3dbc8d3732c0ecfb311d7

HoonBaek:
I think it's good to merge now!

PR #1294, opened 2025-01-08
Add minor util tests

git-babel:
### Changelog
* Added minor test cases for util functions
* Removed unused util functions


src/utils/color.ts
```diff
@@ -18,6 +18,7 @@ export const changeColorToClassName = (color: Colors): string => {
     case Colors.ONBACKGROUND_3: return 'sendbird-color--onbackground-3';
     case Colors.ONBACKGROUND_4: return 'sendbird-color--onbackground-4';
     case Colors.ONCONTENT_1: return 'sendbird-color--oncontent-1';
+    case Colors.ONCONTENT_2: return 'sendbird-color--oncontent-2';
```
HoonBaek, line 21: If we use the ONCONTENT_2, we need CSS set too. But it seems there's no CSS files for this className.

src/utils/useDidMountEffect.ts
```diff
@@ -1,12 +1,13 @@
 import { useEffect, useState } from 'react';
 
 const useDidMountEffect = (func: () => void, deps: Array<unknown>): void => {
-  const [didMount, setDidmount] = useState(false);
+  const [didMount, setDidMount] = useState(false);
   useEffect(() => {
+    console.log('effect');
```
HoonBaek, line 6: Do we really need console.log here?

src/utils/color.ts
git-babel: As you said, we have a logic to convert the enum value to the class name, but there is no actual css for `sendbird-color--oncontent-2`. Should we delete the logic and test for `Colors.ONCONTENT_2`?

HoonBaek: Yes, I think it's good to remove it. We can always add it back when needed.

git-babel: Removed.

HoonBaek (approved):

PR #1444, opened 2026-07-17
[fix] Resolved TypingIndicator.tsx bug not cleaning up newHandlerId

rob-ryszewski:
# Summary

`TypingIndicator.tsx` has a `useEffect` that does:
```
const newHandlerId = uuidv4();
sb.groupChannel.addGroupChannelHandler(newHandlerId, handler);
setHandlerId(newHandlerId);
```

Yet in that `useEffect`'s cleanup function, it targets removing `handlerId` instead of `newHandlerId`:
```
sb.groupChannel.removeGroupChannelHandler(handlerId);
```

This causes `TypingIndicator` to leak group channel handlers that stay subscribed and un-removed.

I came across this in a scenario where this compounded and caused many leaked handlers that then led to other issues.

# Fix
```
sb.groupChannel.removeGroupChannelHandler(newHandlerId ?? handlerId);
```

To defensively cover the scenario were `sb.groupChannel.addGroupChannelHandler` is not defined, and the code never instantiates a `newHandlerId`, it falls back to `?? handlerId`.

### Checklist

### DIFF
--- a/src/modules/GroupChannel/components/TypingIndicator.tsx
+++ b/src/modules/GroupChannel/components/TypingIndicator.tsx
@@ -44,9 +44,10 @@ export const TypingIndicator = ({ channelUrl }: TypingIndicatorProps) => {
   const [typingMembers, setTypingMembers] = useState<Member[]>([]);
 
   useEffect(() => {
+    let newHandlerId;
     if (sb?.groupChannel?.addGroupChannelHandler) {
       sb.groupChannel.removeGroupChannelHandler(handlerId);
-      const newHandlerId = uuidv4();
+      newHandlerId = uuidv4();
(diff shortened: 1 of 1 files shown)


rob-ryszewski:
@sf-tyler-jeong could I get your review? 3 line bugfix to stop leaking handlers, potentially unboundedly

sf-tyler-jeong:
> @sf-tyler-jeong could I get your review? 3 line bugfix to stop leaking handlers, potentially unboundedly

Hi @rob-ryszewski, thanks for the report and the fix! We'll handle this internally through our tracking and release process, and we'll close this PR once the fix is merged. Really appreciate the contribution!

PR #1448, opened 2026-07-29
[CLNP-8774] fix: prevent TypingIndicator from leaking group channel handlers on unmount

sf-tyler-jeong:
## What & why
`TypingIndicator`'s `useEffect` cleanup removed the stale `handlerId` state
instead of the `newHandlerId` that was actually registered via
`addGroupChannelHandler`. The live `GroupChannelHandler` was therefore never
removed on unmount and leaked, accumulating across mount/unmount cycles (e.g.
repeated channel navigation). Originally reported in #1444.

## Changes
- Hoist `newHandlerId` into the effect scope so the cleanup removes the exact
  id it registered, falling back to `handlerId` when no handler was added.
- Add a regression test covering unmount and channel-switch cleanup (fails on
  the previous code, passes with the fix).
- `Channel/components/TypingIndicator` re-exports this component, so both
  module entry points are covered by the one change.

### DIFF
--- a/src/modules/GroupChannel/components/TypingIndicator.tsx
+++ b/src/modules/GroupChannel/components/TypingIndicator.tsx
@@ -44,9 +44,10 @@ export const TypingIndicator = ({ channelUrl }: TypingIndicatorProps) => {
   const [typingMembers, setTypingMembers] = useState<Member[]>([]);
 
   useEffect(() => {
+    let newHandlerId: string | undefined;
     if (sb?.groupChannel?.addGroupChannelHandler) {
       sb.groupChannel.removeGroupChannelHandler(handlerId);
-      const newHandlerId = uuidv4();
+      newHandlerId = uuidv4();
       const handler = new GroupChannelHandler({
         onTypingStatusUpdated: (groupChannel) => {
           // there is a possible warning in here - setState called after unmount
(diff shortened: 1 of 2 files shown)


src/modules/GroupChannel/components/TypingIndicator.tsx
```diff
@@ -44,9 +44,10 @@ export const TypingIndicator = ({ channelUrl }: TypingIndicatorProps) => {
   const [typingMembers, setTypingMembers] = useState<Member[]>([]);
 
   useEffect(() => {
+    let newHandlerId: string | undefined;
     if (sb?.groupChannel?.addGroupChannelHandler) {
       sb.groupChannel.removeGroupChannelHandler(handlerId);
```
danney-chun: handlerId는 더 이상 사용되지 않는 것 같습니다.
최초에 생성될 때 uuid로 생성되고, unmount 시에는 cleanup에서 생성된 ID로 handler를 remove하기 때문에 handlerId가 의미가 없어 보입니다.

src/modules/GroupChannel/components/TypingIndicator.tsx
danney-chun: 이 부분도 더 이상 state로 handler를 관리하는 게 의미가 없어 보입니다.

src/modules/GroupChannel/components/TypingIndicator.tsx
sf-tyler-jeong: 수정했습니다. 감사합니다!

src/modules/GroupChannel/components/TypingIndicator.tsx
sf-tyler-jeong: 수정했습니다. 감사합니다!

danney-chun (approved):
lgtm~

PR #1454, opened 2026-08-26
[fix]: Stop refreshing the group channel collection on mount

danney-chun:
Stop refreshing the group channel collection on mount.

`useGroupChannelList` already creates the initial collection from `channelListQueryParams`, but the effect watching those params used a serialized string as its dependency, so it also ran on the first render and called `refresh()`. That discarded the collection whose first page was still in flight and started over.

While the replacement request was in flight the list was briefly reported as loaded and empty, so the empty-list placeholder rendered for about one round trip before the channels appeared. Apps running with the local cache enabled did not see it, because the replacement collection was already populated from the local database.

- Use the existing `useDidMountEffect` so the refresh runs only when `channelListQueryParams` actually change. A `useRef` guard would not do: React StrictMode double-invokes effects on the same instance, so the second setup would pass the guard and reintroduce the duplicate initialization in development.
- Drop one redundant collection and one redundant channel list request per mount. This happened on every mount regardless of the cache setting.

### Behavior changes

- The overlap used to report the list as loaded while it was still empty, so the auto-select effect called `onChannelSelect(null)` on mount. It now selects the first channel.
- `GroupChannelListUIView` is untouched. The empty-list placeholder still renders when the list is genuinely empty after loading.

### Tests

- `GroupChannelListProvider.queryParams.spec.tsx` — no refresh on mount, including under StrictMode; refresh on a real params change; no refresh when the params keep the same value.
- `GroupChannelListUI.emptyPlaceholder.integration.test.tsx` — renders the provider and the list UI against a fake SDK whose first page stays in flight, and asserts the loading placeholder shows while the empty-list placeholder never does. Reverting this change makes it fail on the rendered placeholder element.

All added tests were confirmed to fail without the source change.

### Follow-up

This removes the refresh that no consumer asked for. Overlapping refreshes are a separate matter and are tracked on their own: a `channelListQueryParams` change or an explicit `refresh()` that lands while a load is still in flight can still momentarily empty the list, and so can a `refresh()` overlapping a scroll-triggered `loadMore()`. None of those are introduced or widened here.

Fixes [CLNP-8827](https://sendbird.atlassian.net/browse/CLNP-8827)

### Changelogs

- Fix the group channel list briefly rendering the empty-list placeholder before channels load when the local cache is disabled

### Checklist

- [ ] Public components / utils / props are appropriately exported

[CLNP-8827]: https://sendbird.atlassian.net/browse/CLNP-8827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/GroupChannelList/components/GroupChannelListUI/__tests__/GroupChannelListUI.emptyPlaceholder.integration.test.tsx
+++ b/src/modules/GroupChannelList/components/GroupChannelListUI/__tests__/GroupChannelListUI.emptyPlaceholder.integration.test.tsx
@@ -1,0 +1,129 @@
+import React from 'react';
+import type { GroupChannel } from '@sendbird/chat/groupChannel';
+import { act, render, screen } from '@testing-library/react';
+
+import { LocalizationContext } from '../../../../../lib/LocalizationContext';
+import { GroupChannelListProvider } from '../../../context/GroupChannelListProvider';
+import GroupChannelListUI from '../index';
+
+type Deferred<T> = {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+};
+
+const createDeferred = <T, >(): Deferred<T> => {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((promiseResolve) => {
+    resolve = promiseResolve;
+  });
+
+  return { promise, resolve };
+};
+
+const createControlledCollection = () => {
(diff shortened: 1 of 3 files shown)


danney-chun:
@codex review

sf-tyler-jeong (approved):
LGTM~

PR #1455, opened 2026-08-27
[fix]: Show the delivery status of messages that never reached the server

danney-chun:
Outgoing messages that never settled could render with no timestamp, no spinner and no error icon at all, so an agent had no way to tell that a message had not been sent. Four defects contributed, each of which is real on its own.

**Status was gated purely on grouping.** `MessageContent` and `ThreadListItemContent` dropped the entire status block whenever `chainBottom` was true. UIKit's own grouping never chains a `pending`/`failed` message, but `chainTop`/`chainBottom` are public props, so an application with a custom `MessageList` computes them itself and bypasses that guarantee. The components now honour the invariant themselves and always render the status of an unsettled message.

**An unmapped status borrowed another status' signal.** `MessageStatus` fell back to the `ERROR` icon for a status it had no icon for, and attached the `sendbird-message-status--sent` hook class on everything that was not `FAILED` — including `NONE`. An application that hides that class to drop the read receipts therefore also hid the bogus error icon, leaving the message completely blank. The icon is now rendered only when the status maps to one, and the hook class follows `isSentStatus`.

**Read receipts were consulted before delivery was confirmed.** `getOutgoingMessageState` ran the group-channel unread/undelivered checks ahead of the `succeeded` check. Those counters short-circuit to `0` for super, broadcast and exclusive channels and for channels with no other joined member, so a message that never left the device could be reported as `READ`, complete with a send timestamp. Delivery is now confirmed first.

**Undelivered messages are no longer chained.** `compareMessagesForGrouping` excluded `pending` and `failed` from grouping; `canceled` and `scheduled` now join them. Without this, collapsing those states to `NONE` above would have started chaining two same-minute canceled messages that previously stayed apart because their read receipts happened to differ.

Every `succeeded` path is unchanged, in both group and open channels. No public signature changes.

## Scope

This addresses the display of messages that were not delivered. It does not change delivery itself.

It covers the status rendering of ordinary outgoing messages in `MessageContent` and `ThreadListItemContent`. Server-originated template messages render through `MessageContentForTemplateMessage`, which has no status path; UIKit has no send path that attaches a template payload, so this is out of scope by design.

## Behaviour changes to be aware of

These are fixes rather than contract changes, but they are visible to integrators.

- Applications passing `chainBottom` from a custom `MessageList` will now see a spinner or an error icon appear on grouped unsettled messages. Layout may shift accordingly.
- `.sendbird-message-status--sent` no longer matches an unmapped (`NONE`) status.
- The `.sendbird-message-status__icon` node is no longer rendered for an unmapped status. The `sendbird-message-status` container remains, but stylesheets or selectors that assume the inner node always exists will be affected.
- In group channels, `canceled` and `scheduled` messages are no longer chained, so their sender header and spacing change.
- In super and broadcast channels, an undelivered message no longer shows a send timestamp.

OpenChannel behaviour is unchanged: `compareMessagesForGrouping`'s no-channel path is untouched.

## Verification

- Full suite: 169 files, 974 tests passing
- Type check, lint and build pass
- No snapshot changes — every existing render path is byte-identical
- Public type declarations compared against the build output; `getOutgoingMessageState`, `compareMessagesForGrouping`, `isSameGroup` and `OutgoingMessageStates` are unchanged
- Rendered a status matrix (6 states x light/dark x with and without an application hiding `--sent`) in Chrome and compared against the baseline build; the five settled states are pixel-identical

Fixes [CLNP-8803](https://sendbird.atlassian.net/browse/CLNP-8803)

### Changelogs

- Fixed a bug where an outgoing message that failed to send could render without any status indicator when a custom message list grouped it
- Fixed a bug where a message with no mapped status showed an error icon and carried the sent-status class
- Fixed a bug where an undelivered message could be reported as read in super and broadcast channels
- Fixed a bug where canceled and scheduled messages could be grouped with adjacent messages

## Checklist

- [x] **Public components / utils / props are appropriately exported**

---

Independent correctness review by Codex (two rounds).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[CLNP-8803]: https://sendbird.atlassian.net/browse/CLNP-8803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Thread/components/ThreadList/ThreadListItemContent.tsx
+++ b/src/modules/Thread/components/ThreadList/ThreadListItemContent.tsx
@@ -11,8 +11,11 @@ import { MessageEmojiMenu } from '../../../../ui/MessageItemReactionMenu';
 import Label, { LabelTypography, LabelColors } from '../../../../ui/Label';
 import {
   getClassName,
+  isFailedMessage,
   isMultipleFilesMessage,
   isOGMessage,
+  isPendingMessage,
+  isSendableMessage,
   isThumbnailMessage,
   SendableMessageType,
 } from '../../../../utils';
@@ -122,6 +125,11 @@ export default function ThreadListItemContent(props: ThreadListItemContentProps)
   const isByMe = (userId === (message as SendableMessageType)?.sender?.userId)
     || ((message as SendableMessageType)?.sendingStatus === 'pending')
     || ((message as SendableMessageType)?.sendingStatus === 'failed');
+  // A message that has not settled must always show its status. chainTop/chainBottom are
+  // public props, so a custom list can chain one and silently remove the only indication
+  // that the message never went out.
+  const isUnsettled = isSendableMessage(message)
+    && (isPendingMessage(message) || isFailedMessage(message));
   const useReplying = !!((replyType === 'QUOTE_REPLY' || replyType === 'THREAD')
     && message?.parentMessageId && message?.parentMessage
     && !disableQuoteMessage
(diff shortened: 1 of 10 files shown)


danney-chun:
@codex review

sf-tyler-jeong (approved):
LGTM~

PR #1464, opened 2026-09-10
[CLNP-8898][fix]: scroll open channel to bottom only for its own sent messages

sf-tyler-jeong:
## Summary

An open channel unexpectedly scrolls to the bottom whenever a message is sent in a **different** channel that shares the same `SendbirdProvider`. A user who has scrolled up to read history loses their position.

## Root cause

`OpenChannelProvider`'s pubSub subscribers for `SEND_USER_MESSAGE` and `SEND_FILE_MESSAGE` called `scrollIntoLast()` **before** (and outside) the `channel && channelUrl === channel.url` guard. The pubSub instance is shared across every module under a single `SendbirdProvider`, and GroupChannel / Channel / Thread sends all publish these topics to it. As a result, a send in any other channel triggered the open channel's subscriber and scrolled its list to the bottom. Only the scroll leaked — the message dispatch was already gated by the channel guard.

## Fix

Move `scrollIntoLast()` inside the channel-URL guard for both subscribers, so an open channel scrolls only for its own sends — matching how the GroupChannel / Channel modules already behave. Sending in the current open channel still scrolls to the bottom as before.

## Tests

Added provider-level regression tests (`OpenChannelProvider.spec.tsx`, 8 cases):
- same-channel user / file message send scrolls to the bottom
- cross-channel user / file message send does not scroll
- payload without a channel does not scroll
- `SEND_MESSAGE_START` never scrolls
- subscribers detach on unmount
- two providers under one `SendbirdProvider`: a send in one channel scrolls only that channel (asserts the exact scroll ref)

Proven to fail on the pre-fix code (4/8) and pass after the fix (8/8). Also manually verified in the testing app.

## Backward compatibility

No public API, prop, export, type, or `StringSet` changes — internal behavior fix only.

### DIFF
--- a/src/modules/OpenChannel/context/OpenChannelProvider.tsx
+++ b/src/modules/OpenChannel/context/OpenChannelProvider.tsx
@@ -211,8 +211,8 @@ const OpenChannelProvider: React.FC<OpenChannelProviderProps> = (props: OpenChan
     }
     subscriber.set(topics.SEND_USER_MESSAGE, pubSub.subscribe(topics.SEND_USER_MESSAGE, (msg) => {
       const { channel, message } = msg;
-      scrollIntoLast(0, conversationScrollRef);
       if (channel && (channelUrl === channel?.url)) {
+        scrollIntoLast(0, conversationScrollRef);
         messagesDispatcher({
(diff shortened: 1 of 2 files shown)


danney-chun (approved):
lgtm~

